PR #202, opened 2014-02-28
First stab at implementing automatic loading of more lines when you scroll up

torhve:
This PR depends on the nomore branch since it has to know when to not request more lines.


lorenzhs:
This is a year old, I'll close it for now. Still want to have that feature at some point though ;)

PR #476, opened 2014-10-16
Cancel all outstanding notifications when disconnecting / navigating away

### LINKED ISSUE #296: Cancel outstanding notifications when disconnecting / navigating away / closing page
When disconnecting, we should cancel all outstanding notifications, and also when navigating away without "disconnecting" (I'm putting that in quotes because of #242 or another bug with the same effect, which results in WeeChat having closed the connection but GB not properly noticing that, thinking it's still connected).

lorenzhs:
Fixes #296

If someone could test this, that would be great. Chrome's notification have [a tendency to really fuck up my system](https://code.google.com/p/chromium/issues/detail?id=383837)

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -221,6 +221,11 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     });
 
     $rootScope.$on('relayDisconnect', function() {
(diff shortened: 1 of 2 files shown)


lorenzhs:
Reworked this not to pollute $rootScope

PR #477, opened 2014-10-18
[WIP] Start implementing auto-reconnect

lorenzhs:
I was recently travelling by train through an area with rather flaky mobile coverage, and it sucked to have to reconnect all the time, especially since GB throws you out of your current view and onto the landing page, so you can't even keep reading things that you already have offline. Instead, it would be nice to display a "reconnecting..." banner, and try to reconnect periodically, and allow the user to take up where they left off.

Ideas to take this even further:
- command queue while offline (send this to that buffer)

Things that need to be done before this can be merged:
- The UI is currently really bad. It covers actual text, and hasn't been tested on mobile
- A "reconnect now" link needs to be added (exponential backoff → timeouts can become quite large)
- Needs to be tested in a range of scenarios

You can try this out by manually closing the connection in WeeChat (/relay → D), or disturbing your internet connection in a way that doesn't run into a timeout (on a phone, that would be switching from mobile network to wifi or the other way around)

TODO:
* [x] reconnect now button
* [ ] make sure we get **all** new lines, which is not always the case (I suspect filtered lines)

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -267,11 +267,13 @@ td.time {
     top: 35px;
     position: fixed;
     z-index: 9999;
-    background-color: #eee;
-    color: #333;
-    width: 100%;
+    width: 80%;
     margin: 0;
     padding: 5px;
+    left: 10%;
+}
+#reconnect a {
+    color: white;
 }
 
 .footer {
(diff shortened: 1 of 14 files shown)


torhve:
I rebased this and added a few tiny fixes. Tested OK on desktop and mobile.

PR #496, opened 2014-11-02
Parse colours in buffer title

### LINKED ISSUE #308: Colours in buffer title (topic) don't render
Glowing Bear does not parse colour codes in the /topic.

Example:

lorenzhs:
Requires a WeeChat version including https://github.com/weechat/weechat/commit/013165209af818daba6b6c5869e822ec9cb86fc3, i.e., WeeChat 1.1 or a November 2014 nightly (or later). Ffor previous versions, see https://github.com/weechat/weechat/issues/237

Fixes #308

### DIFF
--- a/index.html
+++ b/index.html
@@ -215,7 +215,9 @@ $ openssl req -nodes -newkey rsa:4096 -keyout relay.pem -x509 -days 365 -out rel
           </a>
           <button ng-if="debugMode" ng-click="countWatchers()">Count<br />Watchers</button>
         </div>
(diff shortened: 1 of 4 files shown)


lorenzhs:
WeeChat 1.1 was released which fixes the bug mentioned above, so we should be good to go on this. Any opinions?

torhve:
Is there any side effects if you haven't upgraded?

lorenzhs:
Since WeeChat sent IRC colour codes previously, and those don't collide with WeeChat colour codes, I don't expect any side effects. While I haven't tested, I'd be very surprised if something broke

torhve:
I will test it and merge it

PR #518, opened 2014-12-12
Twemoji Support, Take 2

rscullin:
Similar to the PR from before ( #505 ), but now Twemoji parsing is optional and it defaults to disabled. I still plan  on adding UA detection for better default handling, but wanted to get _something_ out while I figure out the best way to do so.


lorenzhs:
Superseded by #530

PR #524, opened 2014-12-22
[PLS TEST] Settings module

lorenzhs:
Tadaaa! Let's unclutter `$rootScope`.


torhve:
I tested and rebased and forced pushed this, looks good to me.

PR #528, opened 2014-12-30
Add some basic tests for filters

lorenzhs:
This depends on #526 and #527. The tests will fail until those are merged.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -3,7 +3,10 @@
 
 var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch']);
 weechat.config(['$compileProvider', function ($compileProvider) {
-    $compileProvider.debugInfoEnabled(false);
+    // hack to determine whether we're executing the tests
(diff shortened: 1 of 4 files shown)


js/glowingbear.js
```diff
@@ -3,7 +3,10 @@
 
 var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch']);
 weechat.config(['$compileProvider', function ($compileProvider) {
-    $compileProvider.debugInfoEnabled(false);
+    // hack to determine whether we're executing the tests
+    if (typeof(it) === "undefined" && typeof(describe) === "undefined") {
```
cormier: I don't think there is a better way to do this in the current situation.
From what I understand the preferred way to do this is to use a per-environment javascript config file that is set at build time, similarly to what is described in this blog post:
http://newtriks.com/2013/11/29/environment-specific-configuration-in-angularjs-using-grunt/

Since one of our requirements is that one should be able to run Glowing Bear directly from the repo (no build), we obviously cannot do this, unless we drop this requirements. I have no problem merging this patch, but I think we should also ask around on the channel and validate that the no-build requirement is still helpful.

lorenzhs: yuck. Thanks for the insight though!

PR #529, opened 2015-01-03
Display ogv, mp4 and webm videos inside a video tag

cormier:
Add a plugin that displays mp4, ogv and webm videos inside a <video> tag

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -268,6 +268,15 @@ plugins.factory('userPlugins', function() {
     });
 
     /*
+     * mp4 video Preview
+     */
+    var videoPlugin = new UrlPlugin('video', function(url) {
+        if (url.match(/\.(mp4|webm|ogv)\b/i)) {
+            return '<video class="embed" width="560"><source src="'+url+'"></source></video>';
+        }
+    });
+
+    /*
      * Cloud Music Embedded Players
      */
     var cloudmusicPlugin = new UrlPlugin('cloud music', function(url) {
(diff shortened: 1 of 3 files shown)


js/plugins.js
```diff
@@ -280,6 +280,25 @@ plugins.factory('userPlugins', function() {
     imagePlugin.name = 'image';
 
     /*
+     * mp4 video Preview
+     */
+    var videoPlugin = new Plugin(function(message) {
+
+        var url = message.match(urlRegexp);
```
lorenzhs: why don't you use `urlPlugin` to easily match multiple videos in one message?

js/plugins.js
```diff
@@ -280,6 +280,25 @@ plugins.factory('userPlugins', function() {
...
+        var content = null;
+
+        if (url) {
+            url = url[0]; /* Actually parse one url per message */
+            if (url.match(/\.(mp4|webm|ogv)$/)) {
+                return '<video width="100%"><source src="'+url+'"></source></video>';
```
lorenzhs: 100%? Doesn't that make the video huuuuge? I think I'd prefer a Reddit Enhancement Suite-like small default size with click&drag to resize. In any case 100% seems like a bad default to me

cormier: Maybe that's a little bit big. I don't have a strong opinion on this, so what would you prefer? 75% 50%?

lorenzhs: Something like youtube (width=560px, height=315px)? I mean, the player has a full-screen button anyway, so we don't need it to be huge.

cormier: Let's specify only the width to keep the aspect ratio. I would rather keep the "width" a proportion to accomodate different browser sizes though.

lorenzhs:
I don't like the `width=75%` part here, that's way too big imho. Why not use the same as youtube, 560x315 (with a `max-width=100%` for mobile)? The browsers' video player has a full screen button anyway should you want to watch it in huge

lorenzhs:
I changed the width thingy while rebasing it, hope you don't mind

PR #530, opened 2015-01-03
Adds emoji support using Twitter's twemoji

lorenzhs:
I addressed mine and @torhve's comments on #518 (and #505) since nothing has happened for three weeks. I also squashed the whole thing into one commit. It was @rscullin who did the real work, though.


lorenzhs:
\o/

PR #531, opened 2015-01-14
plugins: (image) support :size suffix

vivien:
Some images URL (such as Twitter hosting site twimg.com) are suffixed
with :size where size can be small, medium or large.


lorenzhs:
good idea!

PR #535, opened 2015-01-17
Hide Chrome status bar when hovering links in the buffer list

anders:
Removing href="#" works. I don't think this should cause any issues, but maybe there is a reason for the href to be there?

I tested in Chrome, Firefox and Safari on OS X, works fine.


anders:
Chrome, before this patch:
how it looks in Chrome before patch

torhve:
One reason is often to get pointer icon when you hover, is this still there with this patch?

anders:
Yes, it's still there. At least in Chrome and Firefox.

MeltedLux:
<3

PR #536, opened 2015-01-18
Add theme-color meta tag for Android 5

lorenzhs:
@torhve's idea iirc


cormier:
I didn't know we could set this value. That's nice: thanks guys for letting me learn something new everyday! ;-)

PR #538, opened 2015-01-18
[WIP, TAKE IT AND FIX] Update Bootstrap

lorenzhs:
This breaks the accordion, I added a minimal fix of sorts but the animation
is still broken. Should be fixed before merge...


lorenzhs:
Superseded by #780

PR #542, opened 2015-01-31
plugins: add Vine support

vivien:
This commit adds support for Vine embeddable content.


lorenzhs:
This should use urlPlugin...

PR #545, opened 2015-02-04
Request firefox app permissions

torhve:


torhve:
It's just for safety, but I noticed some things didn't work in firefox desktop app without some of these.

lorenzhs:
Ha, weird. Whatever.

PR #547, opened 2015-02-08
Overhaul plugins

lorenzhs:
- Make plugin constructors more convenient to use
- Rewrite many plugins to use UrlPlugin. As a result, youtube URLs without http / https don't work anymore, but everybody copy/pastes those anyway so the protocol is there, imho

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -312,15 +312,21 @@ plugins.factory('userPlugins', function() {
     /*
       * Asciinema plugin
      */
-    var asciinemaPlugin = new Plugin(function(message) {
-
-        var regexp = /^https?:\/\/(www\.)?asciinema.org\/a\/(\d+)/;
-        var match = message.match(regexp);
+    var asciinemaPlugin = new Plugin(urlPlugin(function(url) {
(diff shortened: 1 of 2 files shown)


js/plugins.js
```diff
@@ -247,169 +245,148 @@ plugins.factory('userPlugins', function() {
...
+
+    var tweetPlugin = new UrlPlugin('Tweet', function(url) {
+        var regexp = /^https?:\/\/twitter\.com\/(?:#!\/)?(\w+)\/status(?:es)?\/(\d+)/i;
+        var match = url.match(regexp);
+        if (match) {
+            url = 'https://api.twitter.com/1/statuses/oembed.json?id=' + match[2];
+            return function() {
+                var element = this.getElement();
+                jsonp(url, function(data) {
+                    // sepearate the HTML into content and script tag
```
vivien: Minor, not related to this patchset, to fix later: `s/sepearate/separate/`

lorenzhs: thanks, fixed it

lorenzhs:
Any comments on this?

PR #548, opened 2015-02-08
Fix automatic embedding of asynchronous plugins (jsonp)

### LINKED ISSUE #540: Problem with tweet embedding
Tweets are not properly embedded, and the hiding button looks weird (see screenshot below)

capture d ecran 2015-01-22 a 15 36 51

Additional information:
- I'm using firefox; I tried with Chrome but the problem stays so it does not seem browser specific
- I'm using a mac right now, I haven't tried to reproduce it on another OS yet
- Chosing "Hide embedded content by default" in the parameters seems to solve the problem (but that's not a real solution)

lorenzhs:
Closes #540


lorenzhs:
@cormier @torhve can we get this merged? It's a bugfix that was requested by a user

PR #549, opened 2015-02-08
Improve app installation instructions

### LINKED ISSUE #544: Firefox-based App + Self-signed SSL == :-(
It's really non-obvious how to use the Firefox-based app (Install app -> big blue "Install Firefox app" button). How to do it should be doc'ed and linked to next to the app install button.

Specifically the app doesn't share state with the installed firefox, so you can't just accept the self-signed ssl cert there. And you can't open a normal browser window from within the app from which to accept the cert. So you have to do the following:

(a) In firefox, go to https://machinename:port where machinename and port correspond to the box you have weechat on. No web page will come up, but you should be prompted to accept the self-signed certificate.

(b) Close firefox and glowing bear

(c) Find the firefox profile for the normal firefox in which you accepted the self-signed certificate, and the profile for the glowing bear app. These may be in completly different directory trees. To find this I just searched for a file that's always in profiles; at the time, the "urlclassifierkey3.txt" file fit the bill.

(d) Copy the "cert_override.txt" file from your normal profile to the glowing bear profile. Haven't tested this, but I'm guessing a symbolic link would also work, and have the advantage of easy updates if the machine you have weechat running on changes.

lorenzhs:
Also fix encryption instructions URL for those who proxy their relay

Fixes #544 (most of it, anyway)

### DIFF
--- a/index.html
+++ b/index.html
@@ -158,7 +158,7 @@
           <div id="collapseThree" class="panel-collapse collapse in">
             <div class="panel-body">
               <p>If you check the encryption box, the communication between browser and WeeChat will be encrypted with TLS.</p>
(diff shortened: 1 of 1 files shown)


torhve:
I like this!

Issue #543, opened 2015-01-31
plugin tests need improvements


vivien:
## plugin tests don't support multiple matches

The following line:

> Hey! Check this https://www.youtube.com/watch?v=HO1OV5B_JDw and https://www.youtube.com/watch?v=nVjsGKrE6E8 :-)

will embbed 2 Youtube videos. However, adding this line to the test cases will fail.

## Plugin tests don't check output

Tests must be extended to also check the expected output. If you're regex has multiple matches and you made the mistake to return the bad index (let's say `undefined`), the test will still pass.

For instance, try the following patch:

```` patch
diff --git a/js/plugins.js b/js/plugins.js
index fd9bf51..d364a9b 100644
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -197,7 +197,7 @@ plugins.factory('userPlugins', function() {
 
         // iterate over all matches
         while (match !== null){
-            var token = match[1];
+            var token = match[0];
             var embedurl = "https://www.youtube.com/embed/" + token + "?html5=1&iv_load_policy=3&modestbranding=1&rel=0&showinfo=0";
             content.push('<iframe width="560" height="315" src="'+ embedurl + '" frameborder="0" allowfullscreen frameborder="0"></iframe>');
             // next match
````

## Note on regexp

Plugins must be written in the way that they support multiple contents *per line*.
We might want to impose the following regexp syntax: `/https?:\/\/(www\.)?foobar?/g`
please note the global `g` flag and no line delimiters (`^` or `$`).

Also, the plugin code must be written like `while (match) { content.push(); } return content;` (see the youtube plugin or the Vine plugin).
Maybe this peace of code can be factorized which would allow to only provide the regexp portion to compile and the expected output. For instance:

```` javascript
var vinePlugin = new Plugin("Vine", "vine.co/v/([a-zA-Z0-9]+)(/.*)?", function(matches) {
    var id = matches[0];
    var embedurl = "https://vine.co/v/" + id + "/embed/simple?audio=1";
    return '<iframe class="vine-embed" src="' + embedurl + '" width="600" height="600" frameborder="0"></iframe><script async src="//platform.vine.co/static/scripts/embed.js" charset="utf-8"></script>';
});
````

lorenzhs:
That second thing is exactly why I created urlPlugin, to simplify that! You're right about the first one as well, we need to improve our tests.

lorenzhs:
[...]

PR #551, opened 2015-02-08
Plugins/url pattern

vivien:
This PR mainly factorizes regexps (case insensitivity and protocol prefix) and removes redundant matching code from js/plugins.js. This makes it more convenient to write URL Plugins. Parts have been discussed in #543.

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -272,12 +272,12 @@ plugins.factory('userPlugins', function() {
      */
     var cloudmusicPlugin = new UrlPlugin('cloud music', function(url) {
         /* SoundCloud http://help.soundcloud.com/customer/portal/articles/247785-what-widgets-can-i-use-from-soundcloud- */
-        if (url.match(/^https?:\/\/soundcloud.com\//)) {
+        if (url.match(/\bsoundcloud.com\//)) {
             return '<iframe width="100%" height="120" scrolling="no" frameborder="no" src="https://w.soundcloud.com/player/?url=' + url + '&amp;color=ff6600&amp;auto_play=false&amp;show_artwork=true"></iframe>';
         }
 
         /* MixCloud */
-        if (url.match(/^https?:\/\/([a-z]+\.)?mixcloud.com\//)) {
+        if (url.match(/\bmixcloud.com\//)) {
             return '<iframe width="480" height="60" src="//www.mixcloud.com/widget/iframe/?feed=' + url + '&mini=1&stylecolor=&hide_artwork=&embed_type=widget_standard&hide_tracklist=1&hide_cover=" frameborder="0"></iframe>';
         }
     });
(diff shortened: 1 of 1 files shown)


lorenzhs:
This is quite cool. I'll have to take a closer look later!

lorenzhs:
As mentioned on IRC, these regex escaping issues are super-annoying, and to prevent mistakes we should add tests that ensure certain things are **not** recognized by the plugins, like the youtube1com/foo example or youtube.com.foobar/baz

lorenzhs:
cherry-picked 7bc0204cd21b869ea61dfd60a401364e53ceb8f0 onto master

lorenzhs:
I tested the following and it works in Chrome and Firefox:
```diff
diff --git a/js/plugins.js b/js/plugins.js
index 423360a..6b11bb5 100644
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -34,7 +34,12 @@ var urlRegexp = /(?:ftp|https?):\/\/\S*[^\s.;,(){}<>]/g;
 var UrlPlugin = function(name, urlPattern, urlCallback) {
     return {
         contentForMessage: function(message) {
-            var regexp = new RegExp(urlPattern, 'i');
+            var regexp;
+            if (urlPattern instanceof RegExp) {
+                regexp = new RegExp(urlPattern.source, 'i');
+            } else {
+                regexp = new RegExp(urlPattern, 'i');
+            }
             var urls = message.match(urlRegexp);
             var content = [];
```

cormier:
@vivien: Since `rewrite-with-urlplugin` has been merged in master, I think that this should be rebased on master as well. Other than this, I find it most excellent and would gladly merge it!

vivien:
@lorenzhs had good points on the review. We could indeed merge this work and continue on top of it if we need parts of it, otherwise I can propose a v2. Unfortunately I'm quite busy these days, but in the big picture, here's what I have in mind:

1. Get rid of regexp in URL plugins. Hash an URL and give it to the plugin. This makes much clearer and less cryptic URL handling. See: https://gist.github.com/vivien/50f329e36d404dae00ab
2. Do not parse URLs from a message for *every* plugin. This means having a single UrlPlugin with a list of *known URL handlers*. See: https://gist.github.com/vivien/500e8abbf8dbfa1efc1e

The only concern with 2 is that I don't know how to return the plugin name.
What do you guys think?

lorenzhs:
I've tried to come up with something easier than the `if` blocks, but the potential conditions are too complex to formulate them as a simple object. The youtube condition would look something like this monstrosity and be even less readable than the if blocks:
```javascript
[
    {
        domain: ["com", "youtube"],
        tokens: [
            {conditions: [{path: 0, value: "watch"}], param: "v"},
            {conditions: [{path: 0, value: "embed"}], path: 1},
        ]
    },
    {
        domain: ["be", "youtube"],
        tokens: [{path: 0}]
    }
]
```
So maybe we should consider multiple (domain, path) pairs of regexes? That way, we could capture parameters a lot better and still not end up with ten lines of URL foobar.
```javascript
var regexes = [
    [/\byoutube\.com/, /^watch?([a-zA-Z0-9]+=[a-zA-Z0-9]+)(&[a-zA-Z0-9]+=[a-zA-Z0-9]+)*/],
    [[/\byoutube\.com/, /^embed/[a-zA-Z0-9]+/],
    [[/\byoutu\.be/, /^[a-zA-Z0-9^]/]]
```
-- but then it becomes a pain to figure out which of the parameters was `v`.

I don't have a good solution :(

lorenzhs:
Was there a reason why we haven't merged this yet? If we added the regex passing as noted in https://github.com/glowing-bear/glowing-bear/pull/551#issuecomment-74389573 I think this would be really nice.

vivien:
That can be a first shot indeed. I'll try to check that asap.

PR #559, opened 2015-03-15
fixes for reconnect feature

torhve:
Reinitalise models, and more apply state


torhve:
Merging this since it unbreaks the reconnect feature and shouldn't cause any other problems.
Also it fixes a bug with bufferlist updating I think.

PR #560, opened 2015-03-20
Use transform to improve smootheness.

torhve:
Also push bufferlines in.


lorenzhs:
Thanks for buttery smoothness. Stand by while I attempt to merge with the Github android app...

PR #563, opened 2015-03-21
Load theme without angular to reduce flicker

lorenzhs:


lorenzhs:
closing this since it doesn't actually help reduce flicker

lorenzhs:
reopening because it does work if we initialise with the dark theme

lorenzhs:
this is broken, lol

PR #566, opened 2015-03-21
Twemoji => Emojione

torhve:


js/inputbar.js
```diff
@@ -23,6 +23,11 @@ weechat.directive('inputBar', function() {
                              IrcUtils,
                              settings) {
 
+            // E.g. Turn :smile: into the unicode equivalent
+            $scope.inputChanged = function(a,b,c) {
```
lorenzhs: what's with a, b, c?

PR #568, opened 2015-03-23
Allow td.prefix to grow up to 200px and then wrap.

torhve:
Tested in firefox + mobile, chrome + mobile, IE.


torhve:
This PR needs some work still, it wraps too many lines that aren't words, like if you have spaces in prefix.
Need a workaround for that

torhve:
Solved this a better way.

PR #573, opened 2015-03-24
Shrink nav icons

stenehall:
* Shrink the nav icons and bear for a better mobile experience.
* Shrink the nav icons on desktop for a better navbar.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -81,10 +81,6 @@ input[type=text], input[type=password], #sendMessage, .badge {
 .glyphicon {
     top: 0; /* Fixes alignment issue in top bar */
 }
-.glyphicon-off {
-    top: 1px; /* Fixes for relative glyphicon size */
-    font-size: 28px;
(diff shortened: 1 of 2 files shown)


lorenzhs:
I find them hard enough to hit already, so while I haven't tested your changes yet, are you sure this improves the experience?

stenehall:
Ok, experience might have been the wrong word here. For a cleaner look, rather. If people actually find them hard to hit I'd suggest we keep the icons small but actually increase the hit area a little by adding some padding.

torhve:
Agree with Lorenz here, they are too tiny now. But as you say, what about compromise, decrease the size a little but keep the hit area just as big. Then it will look nicer than now but experience will be the same.

torhve:
Merging this for now, we can take a new look at the icon sizes and navbar height when we get scroll to hide.

lorenzhs:
+1 for increasing them a bit again

PR #574, opened 2015-03-24
Add odd/even backgrounds and increased font-size

stenehall:
* Add odd/even backgrounds for channellist
* Increase font-size for channellist


torhve:
Thanks!

lorenzhs:
I think the font size should only be increased for wide screens (>1400px), on a 1080x1920 (portrait) screen it looks terrible

lorenzhs:
Also the padding seems very large on narrow screens as well

PR #575, opened 2015-03-24
Add notification and unread count to topbar

stenehall:
* Add notification and unread count to topbar


torhve:
Would it be possible to use the function call directly to get the counts, instead of introducing new variables? If not, maybe they could have less ambigious name, maybe something along the lines of total_unread_count, or similar

lorenzhs:
@torhve as I see it the benefit of the variables is that angular knows they haven't changed if they didn't, whereas it has to re-evaluate a function to come to the same conclusion

stenehall:
@lorenzhs and the option would be to expose the factory function and that felt very wrong, or of course we could create a new function for it. But that also seemed excessive.

lorenzhs:
yeah, I agree

PR #576, opened 2015-03-24
Update topbar

stenehall:
Remove topic and add channel name for mobile
* Removing topic from topbar.
* Adding channel name to the topbar.

Click on channel name in topbar to show topic


torhve:
This needs to have channel sign in it too, like # or & etc, or else you can't easily tell channels and private messages apart.

PR #579, opened 2015-03-26
[WIP] Fix commands for buffers whose fullname contain whitespace

lorenzhs:
Use pointers if Weechat version is recent enough (1.0+).
Otherwise, not marking stuff as read is probably the lesser evil than crashing weechat...

Untested!

- [ ] figure out a solution for older WeeChat versions


lorenzhs:
thanks for fixing and merging this :)

PR #583, opened 2015-03-26
Put the badges next to the logo

ametaireau:
I'm not sure that's the right way to fix it, but that does the work for me at least.

Capture:

glowing-bear-badges


ametaireau:
Already fixed, closing.

PR #585, opened 2015-03-26
Handle mobile breakpoints correctly

stenehall:


lorenzhs:
As I mentioned in the old commit somwhere, I'm not a fan of the ## prefix. I don't think we should handle it:
- I don't like the look of it, it has a different indentation than the other channels
- It's not in the RFC
- it's not implemented properly, the indentation is wrong:  -- note how it doesn't line up with the query below

I'm all for the title fixes though.

stenehall:
There we go! All double hashes removed :)

torhve:
I'll wait for @lorenzhs feedback before merging this.

lorenzhs:
Looks good, but can you squash the two commits into one please?

lorenzhs:
just noticed that there is a part in there that is superfluous, `activeBuffer().active` is a tautology ;) we should remove that sometime (hasn't got anything to do with *this* PR)

PR #541, opened 2015-01-27
Added momentum scrolling on overflow elements, for iOS.

tditlu:

PR #586, opened 2015-03-27
Add scrolling momentum to scrollable elements for mobile webkit

rscullin:
This is a tweak to PR #541(adds scrolling momentum to scrollable elements), as that PR seems to cause undesired scrolling issues (touches arbitrarily get absorbed; scrolling is sometimes impossible.)

I've tested this on two iOS devices running iOS 8.x and haven't encountered any scrolling issues.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -145,6 +145,7 @@ input[type=text], input[type=password], #sendMessage, .badge {
     height: 100%;
     overflow-y: auto;
     overflow-x: hidden;
+    -webkit-overflow-scrolling: touch;
     padding-top: 35px; /* topbar */
     padding-bottom: 1px; /* need to force a padding here */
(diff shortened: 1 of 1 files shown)


rscullin:
So, after more testing with the help of #glowing-bear, turns out that this breaks the nicklist entirely in Mobile Safari. I'm going to close this until I (or someone else) has more time to investigate a workaround. It's unfortunately not as simple as just adding a CSS attribute :disappointed: .

lorenzhs:
This may not be the fault of momentum scrolling or iOS, I've experienced nicklist alignment issues on my phone too since we switched to CSS transforms for the bufferlist swiping (PR #560). If you can reproduce the issues you've experienced on master (https://latest.glowing-bear.org/) then I'll gladly reopen and merge this :)

PR #588, opened 2015-03-27
Better disconnect responsiveness

torhve:
Instead of waiting for the websocket to really close, which can take a
long time because network latency, weechat inresponsiveness etc, we just
set our status to disconnected when user wants to disconnect, and we let
the websocket handle the close in its own time in the background.

If the user wants to reconnect this means there will be a new websocket
connection before the old one has failed, but this works just fine.

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -130,17 +130,14 @@ weechat.factory('connection',
         };
 
         var handleClose = function (evt) {
-            ngWebsockets.failCallbacks('disconnection');
-            $rootScope.connected = false;
-            $rootScope.$emit('relayDisconnect');
             if (ssl && evt && evt.code === 1006) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
very nice! I'll test this soon [TM]

js/connection.js
```diff
@@ -235,8 +232,19 @@ weechat.factory('connection',
     };
 
     var disconnect = function() {
+        $log.info('Disconnecting from relay');
         $rootScope.userdisconnect = true;
         ngWebsockets.send(weeChat.Protocol.formatQuit());
+        // In case the backend doesn't repond we will close from our end
+        var closeTimer = setTimeout(function() {
```
lorenzhs: why do you assign this to a variable?

lorenzhs:
Ah screw that timer variable, I want to use this! ;)

PR #600, opened 2015-04-18
Add ellipsis overflow to buffer list name

torhve:


lorenzhs:
`overflow:hidden` breaks the <1400px layout by increasing height on buffers with an unread count

torhve:
Instead of forcing height I removed the CSS rule that increased the height in the first place. Good find, @lorenzhs

PR #603, opened 2015-04-22
Switch to buffer after issuing /query. Fixes #318

### LINKED ISSUE #318: Automatically focus new buffers requested by the user
When we know that the user initiated opening of a new buffer (i.e. clicked a nick in the nicklist or a #channel link), or when they've typed "/query foo" or "/join #bar", we should automatically focus the new buffer.

The issue here is that we have to link the buffer-open event to the preceding event causing the buffer to open, and that means remembering some kind of state.

torhve:
Save outgoing /query nicks to a list and then check that list when we
get a buffer title rename and then switch to the buffer if the name
matches any nick in the outgoing list. Reason to use title event instead
of buffer opened event is that buffer open event doesn't contain the
short name. That would require parsing and guessing full names and in
general be more complicated (but maybe more correct)

This patch can be further improved in the future to check for existing
buffers and switch to them if they already exist.

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -107,10 +107,6 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         var bufferMessage = message.objects[0].content[0];
         var buffer = new models.Buffer(bufferMessage);
         models.addBuffer(buffer);
-        /* Until we can decide if user asked for this buffer to be opened
-         * or not we will let user click opened buffers.
-        models.setActiveBuffer(buffer.id);
-        */
     };
 
     var handleBufferTitleChanged = function(message) {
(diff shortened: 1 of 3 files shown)


js/handlers.js
```diff
@@ -138,6 +134,16 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         // prefix + fullname, which would happen otherwise). Else, use null so that full_name is used
         old.trimmedName = obj.short_name.replace(/^[#&+]/, '') || (obj.short_name ? ' ' : null);
         old.prefix = ['#', '&', '+'].indexOf(obj.short_name.charAt(0)) >= 0 ? obj.short_name.charAt(0) : '';
+
+        // After a buffer openes we get the name change event from relay protocol
+        // Here we check our outgoing commands that openes a buffer and switch
+        // to it if we find the buffer name it the list
+        if (_.contains(models.outgoingQueries, old.shortName)) {
```
lorenzhs: underscore is making this unnecessarily complicated in this instance:
```js
var position = models.outgoingQueries.indexOf(old.shortName);
if (position >= 0) {
    models.outgoingQueries.splice(position, 1);
    models.setActiveBuffer(old.id);
}
```
less code in plain js! :P

js/inputbar.js
```diff
@@ -100,6 +100,14 @@ weechat.directive('inputBar', function() {
                         ab.clear();
                     }
 
+                    // Cache our queries so when a buffer gets opened we can open in UI
+                    if ($scope.command.substring(0, 7) === '/query ') {
```
lorenzhs: This logic seems unnecessarily fragile, why not something along the lines of
```js
var command = $scope.command.split(' ').splice(0, 1); // or /\s/
var commands = ['/query', '/q'];
var index = commands.indexOf(command);
if (index >= 0) {
    var queryName = $scope.command.substring(commands[index].length + 1);
    models.outgoingQueries.push(queryName);
}
```
that would be easily extendable to '/join' as well.

lorenzhs:
Sorry for being such a pain :P I quite like this feature though

torhve:
Thank you for your feedback, I took some of your notes but changed it slightly.

lorenzhs:
thanks a lot! lgtm

PR #605, opened 2015-04-30
Suppress the 'prefix' (nick) when it's the same as previous line.

20after4:
Instead show unicode "Rightwards Arrow With Hook" in place of the nick.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -746,7 +746,6 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
-    var NickClasses = {"wikibugs": ['collapsed'], 'icinga-wm': ['collapsed']};
     function nickFromBufferline(bufferline) {
         if (!bufferline || !bufferline.prefix) {
             return '';
@@ -758,20 +757,17 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
         return nick;
     }
-    $rootScope.bufferlineClassifier = function(bufferlines, index) {
-        var prevNick = nickFromBufferline(bufferlines[index - 1]);
-        var currNick = nickFromBufferline(bufferlines[index]);
-        var classes = NickClasses[currNick] || {};
+    $rootScope.bufferlineClassifier = function(lines, index) {
+        var prevNick = nickFromBufferline(lines[index - 1]);
+        var currNick = nickFromBufferline(lines[index]);
         var res = {};
-
+        if (!currNick || currNick === '') {
+            return res;
+        }
         if (currNick == prevNick) {
             res['repeated-prefix']=true;
-            if (NickClasses[currNick]){
-                res['collapsed']=true;
-            }
         }
         res['nick-'+currNick]=true;
-
         return res;
     };
 
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -746,6 +746,35 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
+    var NickClasses = {"wikibugs": ['collapsed'], 'icinga-wm': ['collapsed']};
+    function nickFromBufferline(bufferline) {
+        if (!bufferline || !bufferline.prefix) {
+            return '';
+        }
+        var prefix = bufferline.prefix;
+        var nick="";
+        for (var i=0; i < prefix.length; i++) {
+            nick += prefix[i].text;
+        }
+        return nick;
+    }
+    $rootScope.bufferlineClassifier = function(bufferlines, index) {
+        var prevNick = nickFromBufferline(bufferlines[index - 1]);
+        var currNick = nickFromBufferline(bufferlines[index]);
+        var classes = NickClasses[currNick] || {};
+        var res = {};
+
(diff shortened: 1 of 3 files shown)


lorenzhs:
Cool feature! However, the implementation adds significant overhead (watchers) that could be avoided using one-time bindings (the :: syntax that's strewn around the file). This avoids re-evaluating of information that never changes and thus speeds up operation. You can have a look at the time stamp mechanism to see how this can be done using classes. In this case, the classes would just hide / show the correct elements.

20after4:
@lorenzsh: ok I'm new to angular, but I think I understand one-time binding now. Does this look ok? I kinda wanted to avoid the !important on the css rules but I don't know of a straightforward way to avoid it.

lorenzhs:
This doesn't work, sometimes it replaces the nick with an arrow even if it doesn't match.

lorenzhs:
I suspect this is because `prefix[0].text` is the mode modifier (`+` in this case) for users with a role. You need to assemble it similarly to `buffer.rtext` (https://github.com/glowing-bear/glowing-bear/blob/86e9fb76968cfeeca295caaf0d9dccac7a885825/js/models.js#L343)

lorenzhs:
also, doesn't look nearly as nice on mobile. rather confusing, actually.

20after4:
lorenzhs: the issue of incorrect classification is fixed. I don't know about the mobile view, I'll try it on my ipad.

js/glowingbear.js
```diff
@@ -746,6 +746,31 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
+    function nickFromBufferline(bufferline) {
```
lorenzhs: this should be using the existing models from models.js. Why are you duplicating this?

js/glowingbear.js
```diff
@@ -746,6 +746,31 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+            return '';
+        }
+        var prefix = bufferline.prefix;
+        var nick="";
+        for (var i=0; i < prefix.length; i++) {
+            nick += prefix[i].text;
+        }
+        return nick;
+    }
+    $rootScope.bufferlineClassifier = function(lines, index) {
```
lorenzhs: you're passing the entire buffer contents to a function for each line. That's terribly inefficient. you could just use `models.activeBuffer()`.

js/glowingbear.js
20after4: I didn't duplicate anything that I know of. Maybe this function belongs in models.js but it doesn't duplicate functionality from there.

js/glowingbear.js
20after4: why is this incredibly inefficient? it's passed by reference, isn't it? I don't iterate over the entire lines[] array

js/glowingbear.js
lorenzhs: Well, in any case `nick` should be an attribute of the bufferline (and it's constructed just like like `rtext`).

js/glowingbear.js
lorenzhs: Hm, true, it's passed by reference. That doesn't mean that it's necessary or a good idea. And yes you can do that with `activeBuffer`. Your `lines` variable refers to the same thing as `models.getActiveBuffer().lines`. What this should look like is something like this:
```html
<tr ng-repeat="bufferline in bufferlines" class="bufferline" ng-class="::{'repeated-prefix': !!bufferline.nick && bufferline.nick == bufferlines[$index-1].nick}">
```
or something like that. No need for a bufferlineClassifier function polluting `$rootScope`.

Maffblaster:
:+1:

yuvipanda:
Is this still considered for merging? I can help rebase if needed.

torhve:
Yeah, if the comments from @lorenzhs are addressed

20after4:
I'll try to fix conflicts and address any issues

lorenzhs:
You can do this now by adding the following to your custom CSS:
```css
.repeated-prefix:after { content:"↪"; }
.repeated-prefix span {display: none; }
```
So I'm closing this

Vaesper:
Reworked the above a little. May be included in Base16 theme in the future, but for now I thought I should share it here as a snippet. Doesn't break highlights and keeps the (hidden) prefix selectable for quoting :) :
```css
.repeated-prefix span {
    position: relative;
    left: -1000px;
}

.repeated-prefix span.hidden-bracket {
    position: absolute;
    left: -1000px;
}

.repeated-prefix span:after {
    content:"↪";
    position: relative;
    left: 1000px;
}

.repeated-prefix span.highlight {
    position: initial;
    left: initial;
}

.repeated-prefix span.hidden-bracket:after,
.repeated-prefix span.highlight:after {
    content: initial;
    position: initial;
    left: initial;
}
```

lorenzhs:
Ooooh including the nick colours, that's fancy!

20after4:
Thanks @lorenzhs  and @Vaesper ...

This is really cool!

20after4:
@Vaesper: I added your example to [wiki/CSS-Customizations](https://github.com/glowing-bear/glowing-bear/wiki/CSS-Customizations)

Vaesper:
:) I added lorenzhs' fix.

PR #609, opened 2015-05-11
Add hidden brackets around user nicks

blasphemy:
What this change does is add <> around user nicks when they are copied. They are not shown in the chat window.

Example:
20:09	@kyle	uh would you please ask for permission before highlighting me? Thanks :)
turns into
20:09	\<@kyle\>	uh would you please ask for permission before highlighting me? Thanks :)

Not sure if there's a better way to do this, but it works for me.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -15,6 +15,10 @@ a {
     cursor: pointer;
 }
 
+.hidden-bracket {
+    font-size: 0px;
+}
+
 td.prefix {
     text-align: right;
     vertical-align: top;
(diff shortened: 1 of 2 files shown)


lorenzhs:
Great idea! Our copy and paste behaviour is terrible at the moment, improvements to that are extremely welcome. Unfortunately, with the current CSS, this messes up line heights for me. What does work for me is something like this:

```css
.hidden-bracket {
  width: 0px;
  color: rgba(0,0,0,0);
  display: inline-block;
}
```

lorenzhs:
actually, could you add spaces to the hidden brackets as well? I don't get the spaces around the nick when copying in Chrome.

blasphemy:
As I mentioned on IRC, I tried @lorenzhs suggested css change, and it broke it in Chrome Dev and FIrefox stable. Anyone have any other ideas?

Issue #606, opened 2015-04-30
Firefox App Installation on Ubuntu Broken 


levlaz:
I installed glowing bear as a Firefox App in Ubuntu. When I tried to launch the application I got a 404 error message saying that the location did not exist. 

Upon further investigation it appears that the `launch_path` argument in `webapp.json` is pointing to `/glowingbear/index.html` which does not really make sense to me. I changed the launch_path variable to be `http://glowing-bear.org` and I am now able to launch the app and use it normally. 

I have only used this on Ubuntu so I am not sure if this is the intended behavior on other systems, but if so then there should be a separate configuration file for Ubuntu (and I asssume other Linux systems). 

My working webapp.json file now looks like this: 

```json 
{"registryDir":"/home/levlaz/.mozilla/firefox/wf1fnvtn.default","app":{"manifest":{"name":"Glowing Bear","description":"WeeChat Web frontend","launch_path":"http://glowing-bear.org","icons":{"32":"/glowing-bear/assets/img/favicon.png","60":"/glowing-bear/assets/img/glowing_bear_60x60.png","90":"/glowing-bear/assets/img/glowing_bear_90x90.png","128":"/glowing-bear/assets/img/glowing_bear_128x128.png"},"installs_allowed_from":["*"],"developer":{"name":"The Glowing Bear Authors","url":"https://github.com/glowing-bear"},"permissions":{"audio-channel-normal":{"description":"Needed to play this app's audio content on the normal channel"},"audio-channel-content":{"description":"Needed to play this app's audio content on the content channel"},"desktop-notification":{}},"default_locale":"en","version":"0.4.8"},"origin":"http://www.glowing-bear.org","manifestURL":"http://www.glowing-bear.org/manifest.webapp","installOrigin":"http://www.glowing-bear.org","categories":[],"receipts":[],"installTime":1430427497401}}
```

Thanks so much for an awesome app! :)

20after4:
I don't even see a webapp.json. I'm also on ubuntu. Where do you see that file?

I'm having a different problem installing the app on ubuntu - it keeps telling me that reinstall is forbidden.

levlaz:
Hey @20after4 All of the firefox apps are installed in your home directory. The reason why you cannot "reinstall" the app is because the file already exists. 

I am not sure if the name of the file is the same but on my machine it is located in ~/..glowingbear-2f382f7d2fd995866615df7aa3ea739b/webapp.json 

[...]

PR #617, opened 2015-05-28
Fix 404 Error in OS X and Linux

levlaz:
In OS X and Linux when the app is launched, it 404s. Making the launch path the url of glowing-bear fixes this issue.

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -3,10 +3,10 @@
   "description": "WeeChat Web frontend",
   "launch_path": "www.glowing-bear.org/index.html",
   "icons": {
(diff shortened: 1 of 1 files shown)


levlaz:
This fixes https://github.com/glowing-bear/glowing-bear/issues/606

lorenzhs:
This would break https://glowing-bear.github.io/glowing-bear/, which is why we haven't done this.

Maybe adding a symlink (a la `ln -s . glowing-bear`) would solve this. However, that would look confusing.

torhve:
Closing this, as it's only applicable in some cases, not all.

PR #618, opened 2015-06-06
Re-implement click to join functionality.

pelmers:
Clicking channel names to join broke because of angular [issue 9515](https://github.com/angular/angular.js/issues/9515). I reimplement the functionality by defining a custom directive which binds the scope to a DOM element and access it from an onclick handler.

It's still nasty.

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -24,7 +24,7 @@ weechat.filter('toArray', function () {
     };
 });
 
-weechat.filter('irclinky', ['$filter', function($filter) {
+weechat.filter('irclinky', function() {
     return function(text) {
         if (!text) {
             return text;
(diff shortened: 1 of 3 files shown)


torhve:
This seems so elaborate and hacky.
Can't we just set the channelname and network in a attribute like data-channel="irc.freenode.#glowing-bear" and then in the openhandler you just read that attribute instead of trying to access the scope?

lorenzhs:
The problem is not accessing the attribute, the problem is calling the openhandler :(

pelmers:
Yes, perhaps it can be written a little more nicely, but I don't see a way around attaching the scope to the DOM somewhere and accessing it on an onclick.

lorenzhs:
Could we maybe use `$interpolate` or `$compile` in the filter to create actual angular elements and not need all those hacks?

pelmers:
I am not very familiar with angular, but in your linked post it does not look like we can compile/interpolate within the filter.
AFAICT, the filter must return plain text, not a DOM/jquery node, so we cannot compile there; it would have to be done in a directive.

pelmers:
Also I realized that sometimes the anchor is not exactly three levels from <stickyscope>. I'll have to write the code to traverse up the DOM and look for a stickyscope.

pelmers:
Done! I also realized that we only have to attach the $scope in one place, because the app only has one controller, so this is done once at the root and we traverse up the DOM tree to get back to it.

lorenzhs:
Hm, that looks way less hacky than I feared this all would turn out. I'll have a closer look soon. In the meantime, could you rebase onto master please?

pelmers:
updated

lorenzhs:
Lovely! Can you squash the two commits so we don't clutter up the repository with the intermediate version?

PR #625, opened 2015-06-12
Check modifier state and don't trigger on any altgr event

torhve:


lorenzhs:
Chrome on Linux didn't accept AltGr before either, btw. Develop for the web, they said. It'll work everywhere, they said.

Issue #624, opened 2015-06-12
Text not justified / hyphenated


michaelshmitty:
I've noticed that the main content window does not properly move a word to the next line if it doesn't fit entirely on the current line.
Not sure I got the terms in the title right to describe the problem.

Here's an example:
![glowing_bear_hyphenation](https://cloud.githubusercontent.com/assets/114845/8126688/203930b0-10f1-11e5-9338-e97db478981b.png)

This makes reading sometimes unpleasant.

torhve:
We use the following CSS: https://github.com/glowing-bear/glowing-bear/blob/master/css/glowingbear.css#L31-L47

You are probably using a browser that doesn't support break-word and thus uses break-all, and also doesn't support inserting hyphens at the correct places.

The reason why we use break-all is to force never having a horizontal scrollbar.

michaelshmitty:
Thanks for the quick reply.

I'm using Firefox 38.0.5, the latest version.
You're right, it doesn't break correctly. I've tried it in Google Chrome and it works fine. Thanks for pointing out where to look for.
If I have time I'll see if I can figure out a fix for Firefox.
Unless someone tells me that due to the nature of using break-all, this can't be fixed in Firefox.

Cheers!

torhve:
I'm not an expert on this, so it's quite possible there exists better solutions.

We'd all be grateful if you find a nicer solution.

lorenzhs:
Yup, I can just repeat what @torhve said, this whole linebreaking thing is a bit of a mess. Improvements would be much appreciated :)

michaelshmitty:
I've been having a go at this today. I tried several approaches, including the one you have tried here:
https://github.com/glowing-bear/glowing-bear/commit/38c96f21c2ffc940f596e27d462d8341b0197ecf

This looks as if it works, but you'll notice that long unbreakable strings like a url for example will bring back a scrollbar in Firefox (in Chrome it looks good, which I assume you use for testing).
I will issue a PR with a very simple change that seems to work in FF and in Chrome. Try it out, if it breaks something, feel free to ignore it.

lorenzhs:
Is this resolved?

lorenzhs:
Closing for inactivity / lack of feedback

PR #627, opened 2015-06-13
Fixes problem with Firefox word / line breaking

michaelshmitty:
References #624.

Firefox would break long lines in the middle of words, disregarding hyphenation rules resulting in poor readability. This change fixes that behavior while still preventing very long unbreakable strings from showing a horizontal scrollbar.

Tested in Firefox 38.0.5 and Google Chrome 43.0.2357.124 (64-bit)
Note the difference between Firefox and Chrome: Firefox splits words with a hyphen to fit them on the line, Chrome moves the entire word to the next line.
I prefer Chrome's way and I wish Firefox would also just move the word to the next line, but I was unable to fix that. Any fix I tried to come up with resulted in Firefox bringing back the horizontal scrollbar for long strings such as urls.

Anyway, have a look at this fix and let me know if it breaks anything. Check my comment on the issue about the word-wrap approach. It breaks long urls.

Cheers!

Michael

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -34,13 +34,11 @@ td.message {
     padding: 1px 1px 1px 5px;
 
 -ms-word-break: break-all;
-    word-break: break-all;
 
     /* Non standard for webkit */
     word-break: break-word;
 
 -webkit-hyphens: auto;
-   -moz-hyphens: auto;
     -ms-hyphens: auto;
         hyphens: auto;
 


lorenzhs:
I still don't understand this patch, but it seems to work and I couldn't find any breakage, so I'll just merge this now and wait for people to complain. Thanks, and sorry for the delay!

michaelshmitty:
Thanks. I got to this fix by experimenting. I am no expert at all on (cross-browser) CSS, so I welcome anyone with the proper knowledge and experience to fix this the right way.
That said, it solved the problem for me and I found no apparent side effects either.

Cheers!

Michael

torhve:
Horizontal scrollbars are back in firefox, not often, but it happens.

I'll rather take split up lines than horizontal scrollbars. And when firefox fixes word-breaking with hypens it will magically get better.

PR #628, opened 2015-06-13
Added a check for DOM 3 getModifierState before attempting to use it

kode54:
… as some browsers may not have implemented that spec yet (Like stable Safari)


torhve:
Thanks a lot for fixing this.

PR #635, opened 2015-07-10
Handle hidden weechat buffers

angus-g:
This adds a new 'hidden' field to the Buffer model, which stays in sync with weechat's notion of hidden buffers.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -654,9 +654,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
             if (buffer.fullName === "core.weechat") {
                 return true;
             }
-            return buffer.unread > 0 || buffer.notification > 0;
+            return (buffer.unread > 0 || buffer.notification > 0) && !buffer.hidden;
         }
-        return true;
+        return !buffer.hidden;
     };
 
     // Watch model and update show setting when it changes
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -81,6 +81,7 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         buffer.trimmedName = buffer.shortName.replace(/^[#&+]/, '');
         buffer.title = message.title;
         buffer.number = message.number;
+        buffer.hidden = message.hidden;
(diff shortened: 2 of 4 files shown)


js/glowingbear.js
```diff
@@ -654,9 +654,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
             if (buffer.fullName === "core.weechat") {
                 return true;
             }
-            return buffer.unread > 0 || buffer.notification > 0;
+            return (buffer.unread > 0 || buffer.notification > 0) && !buffer.hidden;
         }
-        return true;
+        return !buffer.hidden;
```
angus-g: I'm not sure if this is the best way to handle things. The `hasUnread` function seems a bit hacky in general...

angus-g:
I didn't check whether unread messages to hidden buffers still show up as notifications...

js/handlers.js
```diff
@@ -145,6 +146,20 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
...
+        var old = models.getBuffer(buffer);
+        old.hidden = true;
+    };
+
+    var handleBufferUnhidden = function(message) {
+        var obj = message.objects[0].content[0];
+        var buffer = obj.pointers[0];
+        var old = models.getBuffer(buffer);
+        old.hidden = false;
+    };
```
lorenzhs: bleh, we really need to extract the common part of all of these handlers, the code duplication here isn't nice. Not your fault though ;)

lorenzhs:
@torhve you have tons more weechat knowledge than I do, what do you say?

torhve:
Looks good to me, but I haven't tested it.

lorenzhs:
We should probably also ignore them in the Alt+[0-9] shortcuts, right?

angus-g:
That's true. I think weechat proper just removes the buffer so they're all renumbered. Maybe I'd have to request a resync or something after receiving hide/unhide messages.

lorenzhs:
Am I correct in that we can merge this without any (expected) breakage? We can always add the Alt+[0-9] exceptions later.

PR #640, opened 2015-08-05
travis: upgrade to container-based infrastructure

lorenzhs:
since we don't need sudo, say so and get magic speedups!

see http://docs.travis-ci.com/user/migrating-from-legacy/


lorenzhs:
I'm just gonna merge this, this is totally trivial

PR #644, opened 2015-09-11
Add support for supplying credentials in the URL via the hash.

### LINKED ISSUE #619: Pass settings via URL hash
It would be nice to be able to pass settings via the url hash (e.g. `glowing-bear/#server=foo.bar!port=1337!ssl=true!font=sans!fontsize=42pt!noembed=true!mathjax=true`).

See also the discussion in #611

lorenzhs:
Untested. Fixes #619

### DIFF
--- a/index.html
+++ b/index.html
@@ -40,7 +40,7 @@
     <script type="text/javascript" src="js/plugins.js"></script>
     <script type="text/javascript" src="3rdparty/favico-0.3.5.min.js"></script>
   </head>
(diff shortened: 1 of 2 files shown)


evaryont:
It would be nice to expand this a bit to also include an "export". Perhaps a link at the bottom of the settings window?

lorenzhs:
Yeah, shouldn't be too hard to extend though. Just serialize the entire localstorage dictionary. Do you want to give it a try?

torhve:
It's a start! I would like this to get documented for the users too. Could you submit something for that @evaryont  ?

Issue #648, opened 2015-09-25
Add support for giphy


arnottcr:
I am not sure of the exhaustive list, but I know you support youtube and image file embedded content. Could you add support for the gif cdn service https://giphy.com

lorenzhs:
I've personally never seen this service, but plugins are very easy to add -- would you like to give it a try?

PR #650, opened 2015-09-26
Add support for giphy

arnottcr:
added plugin support for giphy gifs as embedded content and created testcases; in reference to #648.


torhve:
Thanks!

PR #651, opened 2015-09-26
simplify giphy regex matching and use original url for image link

arnottcr:
I realized that my code was a tad confusing, so I cleaned it up, removed all but two str.match() calls, and have the image link to the userfacing url, now tls-ed if not before, not the cdn gif image.

The old testcases should still pass for the new code, so they remain unchanged.


arnottcr:
I realized I pushed some dirty code for the youtube plugin that was yet to be finished: resubmitting pull.

lorenzhs:
you can also force-push to the pull request branch to update it. And if you use feature branches instead of sending pull requests, the chances of messing up are smaller.

arnottcr:
yeah, I just --force pushed; I will try the feature branch idea in the future thanks.

arnottcr:
I also see now that if you push new comments to a repo that it will update the pull request dynamically.

lorenzhs:
Like magic! :tada: :tophat: :rabbit2:

PR #652, opened 2015-09-26
simplify giphy regex matching and use original url for image link

arnottcr:
I realized that my code was a tad confusing, so I cleaned it up, removed all but two str.match() calls, and have the image link to the user-facing url, now https if not sent that way, not the cdn gif image as before.

The old testcases should still pass for the new code, so they remain unchanged.

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -363,8 +363,6 @@ plugins.factory('userPlugins', function() {
         var id = url.match(regex);
         if (id) {
             var src = "https://media.giphy.com/media/" + id[1] + "/giphy.gif";
-            /* upgrade any http url to tls, yeah crypto */
-            if (url.match(/^http:/i)) { url.replace(/^http:/, "https:"); }
             return '<a target="_blank" href="'+url+'"><img class="embed" src="' + src + '"></a>';
         }
     });
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -358,17 +358,14 @@ plugins.factory('userPlugins', function() {
   * sample output: https://media.giphy.com/media/feqkVgjJpYtjy/giphy.gif
   */
     var giphyPlugin = new UrlPlugin('Giphy', function(url) {
-        var regex = /^https?:\/\/giphy.com\/gifs\/.*-.*\/?/i;
-        if (url.match(regex)) {
-            /* if the url does not contain a trailing slash, add it */
-            if (! url.match(/\/$/i)) { url = url + "/"; }
-            /* upgrade any http links to tls, yeah crypto */
-            if (url.match(/^http:/i)) { url.replace(/http:/, "https:"); }
-            /* change the url to refrence the giphy cdn url, not the userfacing webserver */
-            url = url.replace(/\/giphy.com\//i,"/media.giphy.com/");
-            url = url.replace(/\/gifs\/.*-/i,"/media/");
-            url = url + "giphy.gif";
-            return '<a target="_blank" href="'+url+'"><img class="embed" src="' + url + '"></a>';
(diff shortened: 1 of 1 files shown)


js/plugins.js
```diff
@@ -358,17 +358,14 @@ plugins.factory('userPlugins', function() {
...
-            url = url.replace(/\/gifs\/.*-/i,"/media/");
-            url = url + "giphy.gif";
-            return '<a target="_blank" href="'+url+'"><img class="embed" src="' + url + '"></a>';
+        var regex = /^https?:\/\/giphy.com\/gifs\/.*-(.*)\/?/i;
+        // on match, id will contain the entire url in [0] and the giphy id in [1]
+        var id = url.match(regex);
+        if (id) {
+            var src = "https://media.giphy.com/media/" + id[1] + "/giphy.gif";
+            /* upgrade any http url to tls, yeah crypto */
+            if (url.match(/^http:/i)) { url.replace(/^http:/, "https:"); }
```
lorenzhs: I think that https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith is the better fit here

js/plugins.js
```diff
@@ -358,17 +358,14 @@ plugins.factory('userPlugins', function() {
...
+            return '<a target="_blank" href="'+url+'"><img class="embed" src="' + src + '"></a>';
```
lorenzhs: you should close the img tag: ``

lorenzhs:
this looks good, just two very minor comments (see inline)

js/plugins.js
torhve: Not in HTML5.

js/plugins.js
arnottcr: that follows, I have an affinity for regexs, but I agree startsWith() reads better and likely runs better

js/plugins.js
arnottcr: that line was lifted from the image plugin var, so let me know what you would like and I can make a change

arnottcr:
I have made the first change and upon consensus on the second will resubmit the pull request. Again, there should be no change in functionality so this is not an urgent pull.

js/plugins.js
lorenzhs: @torhve I didn't know – what's correct in HTML5? Nothing or `</img>`?

lorenzhs:
I think you forgot to push the change, I can't find it.

js/plugins.js
torhve: is self closing, it doesn't have to be closed.

arnottcr:
yeah, I was waiting for confirmation about the  tags first; should be up now

arnottcr:
I started failing testcases with str.startsWith() [ https://travis-ci.org/glowing-bear/glowing-bear/builds/82575468 ]. Because of this and due to the fact that the image link will now conflict with the irc message's url, I decided to remove the tls upgrade. Let me know your thoughts.

lorenzhs:
Hmm, you can also use `indexOf`: `if (url.indexOf('http:') === 0)`, that should work everywhere

arnottcr:
If you would like I can add that back in, but I realized after I started getting issues that it could be confusing to a user that there is a background tls upgrade.

lorenzhs:
As long as the embed is always loaded over `//` or `https://`, I'm fine with it. I just don't want non-TLS-embeds if glowing bear itself is used with TLS. Rewriting the link URL is a bit dodgy though, yeah. So I think it's good as it is now.

arnottcr:
yes, the embedded content is always fetched over `https://`; I agree about the dodgy-ness, hence the removal. Thanks for the help.

PR #657, opened 2015-10-17
Fix sanitization by constructing elements in Javascript.

alexschneider:
This is still technically a hack, the elements should really be defined in a view with templates, but it fixes the issue with sanitize where it can't sanitize if it can't parse HTML.


lorenzhs:
Thanks, this looks interesting! I'll try to have a look later.

PR #661, opened 2015-10-19
Construct plugin elements async. Fixes #660

### LINKED ISSUE #660: GB now loads external resources even with embedding disabled
angular.element code runs with embedding disabled, and I think that causes DOM element to be constructed and thus images/resources loaded.

torhve:
This commit changes the plugins that uses angular.element to load
resources directly to instead return a function that runs
angular.element. When the browser runs angular.element('') and gets
a src attr it will fetch the content in the src. This makes the
browser running gb automatically fetch all plugin content that gets
linked even if the user didn't request it.

Maybe we could change our plugin architecture to split the match
function and the generator functions so this could be solved more
elegantly.

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -285,13 +285,16 @@ plugins.factory('userPlugins', function() {
                 // TODO strip an existing dl=0 parameter
                 url = url + "?dl=1";
             }
-            var element = angular.element('<a></a>')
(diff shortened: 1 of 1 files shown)


lorenzhs:
wwaaaaaah this is awful. web sux.

PR #665, opened 2015-11-07
Feature: imgur button

lorenzhs:
99% of this is @Bakke's work. I cleaned up the history a bit and changed it to add the URL to the newly uploaded image into the input bar instead of sending it directly.

TODO: drag'n'drop should not send directly, use the insert-into-input-bar functionality there as well

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -84,6 +84,25 @@ input[type=text], input[type=password], #sendMessage {
     margin-bottom: 5px !important;
 }
 
+.btn-send-image {
+    position: relative;
+    overflow: hidden;
+    cursor: pointer;
+}
+
+.imgur-upload {
+    position: absolute;
+    bottom: 0;
+    right: 0;
+    cursor: inherit;
+    font-size: 1000px !important;
+    height: 300px;
+    margin: 0;
+    padding: 0;
+    opacity: 0;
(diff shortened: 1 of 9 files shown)


Bakke:
The Imgur-app is not whitelisted yet, so it is limited to1,250 uploads per day. I have submitted two requests to be white-listed (first one two or three weeks ago), but no answer.

Bakke:
@lorenzhs your changes looks really good! Testet the latest version now :)

Using drag and drop is still sending images directly?

lorenzhs:
Thanks! I didn't even look at the drag and drop code to be honest. I'll check it out and maybe factor out the insertion part into an `insertAtCaret` function. 

About the API, I generated a client ID myself now because @torvhe said you wanted to use an "official" one for GB. 1250 should be fine for now, we don't have *that* many users (I think, we don't really know). But you're right, we should get it whitelisted eventually.

lorenzhs:
I changed the drag'n'drop behaviour to match the button, and also submitted a whitelisting request to imgur. Let's see if/when/how they respond.

zakcrawford:
I'd say this is safe to merge, if you run into rate limit issues lemme know.

lorenzhs:
@zakcrawford cool, do you know of a way to see current usage (X/1250)?

PR #668, opened 2015-11-13
New theme and imgur gifv support

hormigas:
Since blacks are more battery efficient on AMOLED displays, I added a black theme that is just a clone of the dark theme with the background set to black. 

I also made it possible to view gifv images from imgur.


lorenzhs:
Thanks, nice idea! But please take a few minutes to clean this up -- could you split this into two pull requests, one for the theme and one for the gifv support? And please squash the fixup commits, they only pollute history and make it harder to see what's going on. I'd also appreciate it if you could rebase onto master and get rid of the merge commit, we've avoided them so far.

PR #670, opened 2015-11-13
IPv6 address literals won't be surrounded by [ ... ] if it's already there

pmelanson:
simple fix. Tested it on my ipv6 server, was able to connect. Tried "[address]", "address", and "[[address]]". Only the last failed.

Do you want me to add a test case for this somehow? I assume it would be one of these protractor end-to-end tests referenced in the readme, but I think that for a tester to run the test successfully, they would have to have a machine capable of supporting ipv6 which my machine doesn't so it doesn't sound like they're super widespread yet.

-pj2melan


lorenzhs:
Thanks!

PR #671, opened 2015-11-13
adding a black theme to better support AMOLED displays.

hormigas:
This is a copy of the dark.css with a black background.

### DIFF
--- a/css/themes/black.css
+++ b/css/themes/black.css
@@ -1,0 +1,2121 @@
+body {
+    background-color: #000000;
+    color: #ddd;
+}
+
+.form-control {
+    color: #ccc;
+    box-shadow: 0px 1px 0px rgba(255, 255, 255, 0.1), 0px 1px 7px 0px rgba(0, 0, 0, 0.8) inset;
+    background: none repeat scroll 0% 0% rgba(0, 0, 0, 0.3);
+    border: 0px none;
+}
+
+html {
+    background-color: inherit;
+}
+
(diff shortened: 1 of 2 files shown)


torhve:
This should be solved using CSS Imports: https://developer.mozilla.org/en-US/docs/Web/CSS/@import instead of just copying the file.

If I merge this PR we would have to copy that file every time we made a change in the other theme, leading to duplicated work.

hormigas:
I'm honestly unclear how this is much different from having the separate light and dark themes. I do understand that black themes are primarily useful to AMOLED displays, but some people prefer black to grey in general. 

And I believe that using @imports to break the css files apart would cause overhead. 

Just adding it into the mobile layout section would suffice, but it also means that we wouldn't be able to further modify the black theme. 

Could you clarify exactly what you'd ideally want?

lorenzhs:
The light and dark theme have big differences (all the colors are different), so it makes sense to keep them apart. The difference between dark and black are something like 5 lines, and 2115 lines are the same. It doesn't make sense to duplicate all that, and makes maintenance much harder (because now you have to edit two files instead of one).

As to the overhead, the dark theme is loaded anyway when the user first loads GB, so it's cached. No overhead. The overhead of downloading >2000 lines twice is much bigger.

I don't like the idea of making the background black for the mobile version of the dark theme. It wouldn't look good on non-AMOLED screens.

lorenzhs:
This is now deprecated.

PR #672, opened 2015-11-13
adding imgur gifv support to the videoPlugin

hormigas:
This just required a slight hack, because all of the gifv files are webm files (if you change the end of the filename to webm it'll go to the video still). Playback didn't work by just assuming we could play gifv, but changing the file extension to .webm did the trick.


torhve:
Thank you for the improvement, and thank you for splitting up into separate PR.

PR #676, opened 2015-11-20
date change messages are chat_day_change color, multiple messages gerated for diff of 1-5 days

pmelanson:
PR for discussion.

Two changes here,

1. the date change message is now a pretty color (specifically, chat_day_change). To do this, I had to change this one line in `js/weechat.js`. I'll point that out in a line comment.

2. If the difference between the current date and the previous message's date is 1-5 dates, inclusive, then it will generate that many date change messages. If it's outside of that range (for example, negative date ranges) only one date change message is generated. I've attached some screenshots of this behaviour.

http://i.imgur.com/7OGs1jF.png (sorry for imgur, git wasn't uploading it properly. I'm on greyhound bus wifi atm.)

I've also tested changing over a leap day. Well, I tried. But `/print -date 2013-03-10T23:59:59 zuul` prints something that has a prefix of 00:59 and a date of March the 12th. So I gave up. If I'm going to have this much trouble over me not knowing what is local time and what is UTC in my testing, I won't because this isn't critical behaviour. It works 90% of the time around leap days though!

Yeah. thoughts?

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -19,12 +19,39 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         models.closeBuffer(bufferId);
     };
 
+    // inject a fake buffer line for date change
+    var injectDateChangeMessage = function(message, buffer, date) {
+        var content = "Date changed to " + date.toDateString();
+        var line = {
+            buffer: buffer,
+            date: date,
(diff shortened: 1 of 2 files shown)


pmelanson:
woops! didn't mean to have master branch as the target. Closing and reopening.

PR #678, opened 2015-11-24
Add date change markers

lorenzhs:
This adds date change markers dynamically. Thanks for @pmelanson for getting my prototype up to speed.

I can kill the merge commit if you'd like but I don't think it hurts.

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -19,75 +19,69 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         models.closeBuffer(bufferId);
     };
 
-    // inject a fake buffer line for date change
-    var injectDateChangeMessage = function(buffer, old_date, new_date) {
-        var old_date_plus_one = old_date;
-        old_date_plus_one.setDate(old_date.getDate() + 1);
-
-        new_date.setHours(0, 0, 0, 0);
+    // inject a fake buffer line for date change if needed
+    var injectDateChangeMessageIfNeeded = function(buffer, old_date, new_date) {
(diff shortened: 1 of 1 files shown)


torhve:
When all the lines are days old and there hasn't been any new messages, shouldn't the last line be a date change line?

lorenzhs:
Hmm yes that would be a good idea to have. I guess this should go into `handleBufferLineAdded` and check whether the last message was sent today?

torhve:
Will you add this before we merge, or wait til later? I think this is good to go as it is, but would like the feature I propose as now it can be a bit confusing, looking like messages are from today.

pmelanson:
I have a bit of free time, @lorenzhs if you don't have time atm I could have a crack at it now that I know what function to start at.

lorenzhs:
Please feel free to go ahead, @pmelanson! It's always nice to have new contributors, it's more fun that way and more people have more ideas.

pmelanson:
Responding to IRC, for a different message than "Date changed to"

for a change of one day:
`Wednesday Nov 25, 2015` (full length week day, abbrev month, comma before year)

for a change of two days:
`Wednesday Nov 25, 2015 (266 days later)` or
`Wednesday Nov 25, 2015 (266 days earlier)`
(same as date1, but doesn't print what the previous date was)

Thoughts on these?

torhve:
I like this as it is, and think it's good to merge. We can always change the wording later when we have a better feel for it after using it for a bit.

If @lorenzhs also approve of the wording I will merge this (or he can)

lorenzhs:
I think this is good to go

PR #681, opened 2015-11-27
disallow nested links in DOMfilter (fixes #680)

### LINKED ISSUE #680: URLs with a hash should not generate an IRC channel link
If there's an URL like `http://localhost/#foo`, the following markup is generated:

```
<a target="_blank" href="http://localhost/#foo"><span>http://localhost/</span></a><a href="#" onclick="openBuffer('#foo');">#foo</a>
```

Obviously the user won't want to join the `#foo` IRC channel, so this is a bug.

dancek:
Here's an idea. Nested `<a>` tags are disallowed per the HTML spec, and skipping filters that would create them would fix #680. This should not cause a notable performance hit, especially considering there should not be much content inside link tags anyway.

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -61,10 +61,15 @@ weechat.filter('inlinecolour', function() {
 
 // apply a filter to an HTML string's text nodes, and do so with not exceedingly terrible performance
 weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
+    // To prevent nested anchors, we need to know if a filter is going to create them.
+    // Here's a list of names. See #681 for more information.
+    var filtersThatCreateAnchors = ['irclinky'];
+
     return function(text, filter) {
         if (!text || !filter) {
             return text;
         }
+        var createsAnchor = filtersThatCreateAnchors.indexOf(filter) > -1;
 
         var escape_html = function(text) {
             // First, escape entities to prevent escaping issues because it's a bad idea
@@ -92,6 +97,7 @@ weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
                 // as innerHTML causes it to be unescaped.
                 var input = escape_html(node.nodeValue);
                 var value = filterFunction(input, extraArgument, thirdArgument);
+
                 if (value !== input) {
                     // we changed something. create a new node to replace the current one
                     // we could also only add its children but that would probably incur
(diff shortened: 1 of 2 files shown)


lorenzhs:
why don't you add a `createsAnchor` parameter that indicates whether the filter passed to DOMfilter creates <a> tags? Then you wouldn't need the `insideLink` check to see whether it created an anchor. Would be a lot less code.

dancek:
Yes, that makes sense.

dancek:
Oh, but this breaks the extra argument support found on line 79-82. I wonder where they're used.

```js
        // hacky way to pass extra arguments without using .apply, which
        // would require assembling an argument array. PERFORMANCE!!!
        var extraArgument = (arguments.length > 2) ? arguments[2] : null;
        var thirdArgument = (arguments.length > 3) ? arguments[3] : null;
```

lorenzhs:
Yeah, you'd have to move those up to 3 and 4 and add the Boolean second argument

lorenzhs:
The emoji and MathJax filters use it to pass settings iirc

dancek:
I'm suggesting a list of filters that create anchors instead, since `irclinky` is currently the only filter this affects, and needing to add parameters every time a filter is used sounds like a nuisance to me.

js/filters.js
```diff
@@ -114,7 +120,11 @@ weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
             if (node === undefined || node === null) return;
             node = node.firstChild;
             while (node) {
-                var nextNode = process(node);
+                var nextNode;
+                // do not recurse inside links if the filter would create a nested link
+                if (!(createsAnchor && node.tagName == 'A')) {
```
lorenzhs: please use triple `=` for comparison

lorenzhs:
Thanks, this looks really good! Would you mind fixing that one `==` instead of `===` and maybe squashing it into one commit?

js/filters.js
dancek: I get that a lot. Due to [how loose equality works in JS](https://github.com/getify/You-Dont-Know-JS/blob/master/types%20&%20grammar/ch4.md#loose-equals-vs-strict-equals), I prefer it in places where I know it will work correctly. And for example, the null check a couple of lines earlier would IMHO be more readable with loose equals: `(node === undefined || node === null)` vs. `(node == null)`.

But yeah, I'll change that. I'm not here to force my opinions. :)

dancek:
Yeah, I was waiting for everything to be accepted before squashing.

js/filters.js
lorenzhs: I'm sure you know it works correctly, but others may not. Loose equality is so strange in JS that I prefer to avoid it completely. And that check above could be even shorter using `(!node)`

lorenzhs:
Awesome, thanks.

PR #682, opened 2015-11-27
Date change at the end of the buffer when changing buffers, more HR dates

pmelanson:
Two changes here.

First, if you change buffer into a buffer where the last message wasn't the current day, injects a date change message.

Second, changed the format of the names. The new behaviour is contained in the commit message of 9334f44

Thoughts?

I've been messing around with setting a different timezone on my weechat relay, just to make sure the whole timezone thing works. As such, I'm going to make this PR, then maybe stay up until midnight local time and see if it generates a date change message.

Dates are fun.

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -20,7 +20,7 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
     };
 
     // inject a fake buffer line for date change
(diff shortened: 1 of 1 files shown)


torhve:
I like it! Thanks for working on this

pmelanson:
my pleasure!

PR #683, opened 2015-11-28
Request nicklist using ptr instead of fullname

torhve:
Noticed this problem in some of my Matrix buffer, which has spaces in them.

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -282,10 +282,13 @@ weechat.factory('connection',
     };
 
     var requestNicklist = function(bufferId, callback) {
-        bufferId = bufferId || null;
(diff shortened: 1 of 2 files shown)


js/connection.js
```diff
@@ -282,10 +282,13 @@ weechat.factory('connection',
     };
 
     var requestNicklist = function(bufferId, callback) {
-        bufferId = bufferId || null;
+        // Prevent requesting nicklist for all buffers if bufferId is invalid
+        if (!bufferId || bufferId === null || bufferId === undefined) {
```
lorenzhs: isn't this kinda redundant? Aren't `!null` and `!undefined` both true?

PR #684, opened 2015-11-28
made date change msg not generate if one of the dates is time 0

pmelanson:
Behaviour noticed today when switching to my relay.relay.list buffer. Made change, now that message isn't generated.


lorenzhs:
why not just `if (old_date.getTime() < 24*3600)`? First day is really easy to compute

lorenzhs:
But seriously, we should check whether it's a free or formatted buffer

lorenzhs:
This is superseded by #685

PR #689, opened 2015-12-01
fix #688 and provide a unit test

### LINKED ISSUE #688: Lines with a hashtag and link crash GB
Certain lines, like the following:

`JVST SAY YES - Switch Ft. Ephwurd by EPHWURD #np https://soundcloud.com/ephwurd/jvst-say-yes-switch-ft-ephwurd`
or
`#crash http://google.com`

cause the latest GB (e63877a) to lock up and become unresponsive when attempted to be displayed. CPU usage goes to 100% and never seems to recover.

dancek:
Sorry about this. I kind of thought that the original code had a block-local variable, ES6 `let` style. But obviously `nextNode` is hoisted and retains its value, causing infinite loops with the new logic.

Last time I was lucky enough to think of a unit test that doesn't hit this bug, too... The new test doesn't test anything useful but does hang in an infinite loop if the issue exists.

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -120,7 +120,7 @@ weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
             if (node === undefined || node === null) return;
             node = node.firstChild;
             while (node) {
-                var nextNode;
+                var nextNode = null;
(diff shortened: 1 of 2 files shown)


torhve:
Thanks for fixing this this quick!

Issue #511, opened 2014-11-25
Switch buffers with Alt+arrow


blasphemy:
I use this extensively on the CLI client, but it doesn't work in glowing bear. Is this possible to implement?

Maffblaster:
I'd like this feature as well if possible. Since the channels are currently displayed in a vertical manner it might be best to use the ↑ and ↓ keys (up and down).

If there are other hot keys available in GB it would be nice to have a legend for available hotkeys. I currently do not know how to view if any hot keys are available.

rickybrent:
I did a fork of GB with alt-up and alt-down. That works great on my linux PC but not on my chromebook. :) 

It's a bit more complicated than it seems since the visual order shown by glowing bear isn't always the same as the actual order by index; since I was just coding it for me I didn't care too much but if I was going to clean it up and make a pull request that'd be important.

Maffblaster:
It would be nice to not have to leave the keyboard when connecting via the web interface, although search can be done with an Alt+g, I still think this feature would be used. :+1:

20after4:
:heart: This is the biggest WTF in glowing-bear. It's sad that this hasn't been implemented but I looked into it and it turned out to be a lot messier and more difficult than I expected.

lorenzhs:
Fixed with #691

PR #691, opened 2015-12-01
Switch to adjacent buffer with alt+arrow up/down

20after4:
A rough implementation of #511

probably only works in nested bufferlist.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -727,6 +727,18 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         settings.nonicklist = !settings.nonicklist;
     };
 
+    $rootScope.switchToAdjacentBuffer = function(direction) {
+        // direction is +1 for next buffer, -1 for previous buffer
+        var sortedBuffers = _.sortBy($scope.getBuffers(), $rootScope.predicate);
+        var activeBuffer = models.getActiveBuffer();
+        var index = sortedBuffers.indexOf(activeBuffer);
+        if (index >= 0) {
+            var newBuffer = sortedBuffers[index + direction];
+            if (newBuffer) {
+                $scope.setActiveBuffer(newBuffer.id);
+            }
+        }
+    };
 
     $scope.handleSearchBoxKey = function($event) {
         // Support different browser quirks
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -294,6 +294,14 @@ weechat.directive('inputBar', function() {
                     return true;
                 }
 
+                // Alt+Arrow up/down -> switch to prev/next adjacent buffer
+                if ($event.altKey && !$event.ctrlKey && (code === 38 || code === 40)) {
+                    $event.preventDefault();
+                    var direction = code - 39;
+                    $rootScope.switchToAdjacentBuffer(direction);
+                    return true;
+                }
+
                 // Alt+L -> focus on input bar
                 if ($event.altKey && (code === 76 || code === 108)) {
                     $event.preventDefault();
--- a/js/models.js
+++ b/js/models.js
@@ -84,7 +84,7 @@ models.service('models', ['$rootScope', '$filter', function($rootScope, $filter)
         var notification = 0;
         var unread = 0;
         var lastSeen = -1;
-        var serverSortKey = fullName.replace(/^irc\.server\.(\w+)/, "irc.$1");
+        var serverSortKey = fullName.replace(/^irc\.server\.(\w+)/, "irc.$1").toLowerCase();
         // There are two kinds of types: bufferType (free vs formatted) and
         // the kind of type that distinguishes queries from channels etc
         var bufferType = message.type;


js/glowingbear.js
```diff
@@ -727,6 +727,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         settings.nonicklist = !settings.nonicklist;
     };
 
+    $rootScope.switchToAdjacentBuffer = function(direction) {
+        // direction is +1 for next buffer, -1 for previous buffer
+        var sortedBuffers = _.sortBy($scope.getBuffers(), 'fullName');
```
lorenzhs: this should probably mimic what the angular stuff does in the html and use `$rootScope.predicate` for sorting instead of hardcoded `'fullName'`

js/glowingbear.js
```diff
@@ -727,6 +727,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+        var i, buffer;
+        var activeBuffer = models.getActiveBuffer();
+        var bufferKeys = Object.keys(sortedBuffers);
+
+        // Try to find adjacent
+        for (i in sortedBuffers) {
```
lorenzhs: you could do something like this:
```js
var index = sortedBuffers.indexOf(activeBuffer);
if (index >= 0) {
    var newBuffer = sortedBuffers[index + direction];
    if (newBuffer) {
        $scope.setActiveBuffer(newBuffer.id);
    }
}
```

js/glowingbear.js
```diff
@@ -727,6 +727,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+            buffer = sortedBuffers[i];
+            if (buffer === activeBuffer) {
+                var newKey = parseInt(i)+parseInt(direction);
```
lorenzhs: why do you need the `parseInt` here? shouldn't it already be a number?

js/inputbar.js
```diff
@@ -294,6 +294,14 @@ weechat.directive('inputBar', function() {
                     return true;
                 }
 
+                if ($event.altKey && (code === 38 || code === 40)) {
+                    $event.preventDefault();
+                    var direction = code - 39;
+                    console.log('direction:', direction)
```
lorenzhs: please remove the debug logging. I use it a lot myself during development but we should keep it clean in production code (maybe we could use $log with appropriate levels but that's another discussion)

js/inputbar.js
```diff
@@ -294,6 +294,14 @@ weechat.directive('inputBar', function() {
...
+                    $rootScope.switchToAdjacentBuffer(direction)
```
lorenzhs: missing semicolon

lorenzhs:
Thanks for implementing this! I hope my comments make sense and that I didn't make too many mistakes typing out code in a comment box ;) Oh and please squash the commits in the end so we don't clobber up the history.

20after4:
@lorenzhs: thanks for all the help! I squashed the commits and addressed your feedback. And it seems to work nicely for me in nested bufferlist mode as well as with a flat list.

lorenzhs:
Great! Some of the imgur bits seem to have gotten caught up in your last commit though, I'm not sure what exactly went wrong there but we'd like to keep them ;)

js/glowingbear.js
```diff
@@ -727,6 +727,19 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         settings.nonicklist = !settings.nonicklist;
     };
 
+    $rootScope.switchToAdjacentBuffer = function(direction) {
+        // direction is +1 for next buffer, -1 for previous buffer
+        var sortedBuffers = _.sortBy($scope.getBuffers(), $rootScope.predicate);
+        var i, buffer;
```
lorenzhs: those two are now unused

20after4:
Ok that should be ready to land then?

js/inputbar.js
```diff
@@ -294,6 +294,13 @@ weechat.directive('inputBar', function() {
                     return true;
                 }
 
+                if ($event.altKey && (code === 38 || code === 40)) {
```
torhve: I would prefer it if this code also checked that ctrlkey wasn't down. As right now this would also work with ctrl-alt-arrow. In the future we might would want to do something different for that, or this might trigger something else in some other layout that we don't know about. And in general, it's better to be explicit if you mean alt+arrow and *only* alt-arrow, not all the other combinations.

lorenzhs: Good catch! The current state interferes with workspace switching for me. Please check that Ctrl isn't pressed a swell.

20after4: done

lorenzhs:
LGTM, haven't tested it though

lorenzhs:
has a weird bug for me, it goes from `#weechat-offtopic` to `NickServ` to `dcormier` which isn't particularly alphabetical. Maybe the uppercase character is messing with the sort order (because it's a lower ASCII char). I suspect angular's sorting ignores case while underscore's doesnt?

20after4:
That is an odd bug. I don't see it on my end even though I do have some uppercase buffers. hmm

20after4:
@lorenzhs: try it now?

lorenzhs:
awesome, it works now. Thanks a lot!

fuzzy76:
This should probably have been mentioned in the landing page as well.

lorenzhs:
Good idea! I just proposed #749

PR #693, opened 2015-12-05
Use buffer localvars instead of regex hackery

### LINKED ISSUE #589: Use localvar instead of regex hackery on buffer names
This would not only be more clean, but also more reliable.

Places where we're using awful ugliness:
- sort by server functionality (serverSortKey thingy)
- notification titles
- ???

lorenzhs:
Fixes #589

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -248,6 +248,11 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
             // Update indentation status
             old.type = localvars.type;
             old.indent = (['channel', 'private'].indexOf(localvars.type) >= 0);
+            // Update serverSortKey and related variables
+            old.plugin = localvars.plugin;
+            old.server = localvars.server;
+            old.serverSortKey = old.plugin + "." + old.server +
+                (old.type === "server" ? "" :  ("." + old.shortName));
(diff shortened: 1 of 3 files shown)


js/handlers.js
```diff
@@ -248,6 +248,11 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
             // Update indentation status
             old.type = localvars.type;
             old.indent = (['channel', 'private'].indexOf(localvars.type) >= 0);
+            // Update serverSortKey and related variables
+            old.plugin = localvars.plugin;
```
lorenzhs: I don't think `plugin` can ever change and it's also known when the buffer is first created (unlike `server`) so I'm not sure this is necessary

js/models.js
```diff
@@ -346,13 +357,20 @@ models.service('models', ['$rootScope', '$filter', function($rootScope, $filter)
             });
         }
 
+        // flatten content and prefix
         var rtext = "";
         for (var i = 0; i < content.length; ++i) {
             rtext += content[i].text;
         }
 
+        var rprefix = "";
```
lorenzhs: hm actually I'm not sure about this anymore, for most messages this is wasted work

PR #694, opened 2015-12-05
WIP: Don't put hidden <> around non-messages' prefixes

### LINKED ISSUE #616: Don't put invisible <> for copy&paste around non-chat prefixes such as join/quit messages
image

As is obvious, the WeeChat join/part characters are customisable, in this case Glowing-Bear seems to treat them as nicks. Same goes for plugins which just /print to a buffer, with a blank nick.
The former should have no prefix+suffix;
The latter should have no nick whatsoever (so no prefix+suffix);

lorenzhs:
Fixes #616

@torhve what other tags do we need to check for so this could work with e.g. matrix?


lorenzhs:
20:45  <+xt>should probably use notify_message intead
20:45  <+xt>or something
20:45  <+xt>but that right there feels costly
20:45  <+xt>should probably do it in pure js on line add instead of in the template

PR #695, opened 2015-12-08
prevent "zalgo text" from exiting the message/bufferline

dancek:
For a review of what zalgo text is and how it works, see
http://stackoverflow.com/q/6579844/659526 . Suffice to say it abuses
unicode to create a vertical mess that goes over other lines of text.

This here patch adds `overflow: hidden` to two different places (for mobile/desktop). I briefly tested this and nothing seems to break, but given my track record I suggest reviewing before merging. :)

Example of zalgo text

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -29,6 +29,7 @@ td.prefix {
     border-right: 1px solid #444;
 }
 td.message {
+    overflow: hidden;
     vertical-align: top;
     width: 100%;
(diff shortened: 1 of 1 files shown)


lorenzhs:
which browser did you use for that screenshot? I can't reproduce in Chrome and Firefox (i.e., no fix needed for either)

lorenzhs:
Thanks! Didn't notice any breakage either.

PR #701, opened 2015-12-13
#700 added fallback mode for unsupported .toLocaleDateString()

### LINKED ISSUE #700: iOS: date change marker doesn't seem to do locale stuff properly
pj2melan on iirc said: I got the supported() function from the MDN page on toLocaleDateString

Safari on my iPad (iOS 9.1.1, iPad Air 1) is not supported according to that function.

pmelanson:
Fixes #700 for iOS and any other browsers that don't properly support this one method taking arguments.

Fix is: detect if `Date.toLocaleDateString()` supports taking in a locale argument and a format options argument, and if the date formats properly then do the same thing we've been doing. Otherwise, keep track that it's not supported properly.
If it's not supported, fall back to a manual implementation of `en-US`. Yeah. With like lookups in arrays and stuff. It's not ideal, it's not internationalized, and I don't like it. Another solution is include a 3rd party JS libs but that's a lot so I left it as is.

I tried to keep performance impact when toLocaleDateString *is* supported to a minimum.

Want me to add a unit test or something? I tested this code by setting `supports_formatting_date = false` before it goes through the rest of the date change function, and it successfully output the right dates.

Thoughts?

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -761,6 +761,23 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
+    $rootScope.supports_formatting_date = (function() {
+        // function toLocaleDateStringSupportsLocales taken from MDN:
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#Checking_for_support_for_locales_and_options_arguments
+        try {
+            new Date().toLocaleDateString('i');
+        } catch (e) {
+            if (e.name !== 'RangeError') {
+                $log.info("Browser does not support toLocaleDateString()," +
+                          " falling back to en-US");
+            }
+            return e.name === 'RangeError';
+        }
+        $log.info("Browser does not support toLocaleDateString()," +
+                  " falling back to en-US");
+        return false;
+    })();
+
     // Prevent user from accidentally leaving the page
     window.onbeforeunload = function(event) {
 
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -59,8 +59,16 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
 
             var content = "\u001943"; // this colour corresponds to chat_day_change
             // Add day of the week
-            content += new_date.toLocaleDateString(window.navigator.language,
-                                                   {weekday: "long"});
+            if ($rootScope.supports_formatting_date) {
+                content += new_date.toLocaleDateString(window.navigator.language,
+                                                       {weekday: "long"});
+            } else {
+                // Gross code that only does English dates ew gross
+                var dow_to_word = [
+                    "Sunday", "Monday", "Tuesday",
+                    "Wednesday", "Thursday", "Friday", "Saturday"];
+                content += dow_to_word[new_date.getDay()];
+            }
             // if you're testing different date formats,
             // make sure to test different locales such as "en-US",
             // "en-US-u-ca-persian" (which has different weekdays, year 0, and an ERA)
@@ -73,8 +81,20 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
                 extra_date_format.year = "numeric";
             }
             content += " (";
-            content += new_date.toLocaleDateString(window.navigator.language,
-                                                   extra_date_format);
+            if ($rootScope.supports_formatting_date) {
+                content += new_date.toLocaleDateString(window.navigator.language,
+                                                       extra_date_format);
+            } else {
+                // ew ew not more gross code
+                var month_to_word = [
+                    "January", "February", "March", "April",
+                    "May", "June", "July", "August",
+                    "September", "October", "November", "December"];
+                content += month_to_word[new_date.getMonth()] + " " + new_date.getDate().toString();
+                if (extra_date_format.year === "numeric") {
+                    content += ", " + new_date.getFullYear().toString();
+                }
+            }
             // Result should be something like
             // Friday (November 27)
             // or if the year is different,


js/handlers.js
```diff
@@ -29,6 +29,15 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         new_date.setHours(0, 0, 0, 0);
         // Check if the date changed
         if (old_date.valueOf() !== new_date.valueOf()) {
+            var supports_formatting_date = false;
+            var should_be_tuesday = new Date(Date.UTC(2012,11,12,3,12,0))
+                .toLocaleDateString("en-US-u-ca-iso8601", {"weekday":"long"});
```
torhve: this feels incredibly hacky. Also, I don't like that code being ran on each inject, why don't we run something once upon loading of gb?

Also, that code returns wednesday in all my browsers.

pmelanson: The alternative given on [the MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#Checking_for_support_for_locales_and_options_arguments) was to check if the right exception is thrown when giving an invalid index. We could do this, but to me that doesn't guarantee that the date will be formatted correctly. I'm just assuming less about a browser in the wild by checking what I'm actually trying to check.

As for returning Wednesday, I think that might be UTC mixup. If we want to keep it as this code, `new Date(new_date.getTimezoneOffset()*60*1000).toLocaleDateString("en-US-u-ca-iso8601", {"weekday":"long"})'` should return Tuesday. If we want to not keep this code and go with the function that the MDN page has, we can do that as well. (But I can't see anything we're going to do be not hacky).

As for running it once on gb loading, where do you suggest putting that?

pmelanson:
I'll make the `getFullYear` change.

js/handlers.js
torhve: I don't understand at all what you mean by UTC mixup and DST. UTC doesn't have DST. And "mixups" can't happen because UTC is UTC "everywhere". Which means code is just wrong, the way I understand it.

As for init, somewhere in WeechatCtrl, there's already many types of detecting stuff there.

pmelanson: Yeah, the code is wrong as it is.

The mixup was that I was using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC#Description which is slightly different than just using the constructor.

As for DST, I was talking about DST in regards to a potential fix. The fix would be to just find January 1 1970 in local time, then make sure it's a Thursday. One would have to account for DST because getTimezoneOffset changes after DST.

However, an even better line of code than that is `new Date(1970, 0, 1).toLocaleDateString("en-US-u-ca-iso8601", {"weekday":"long"})` which should return Thursday. Which turns out is different than `new Date("1970-1-1")`, which constructs January 1st, 1970 in UTC time. `new Date(1970, 0, 1)` constructs it in local time.

Thoughts?

torhve: At this point, why not just use exceptions instead? This seems all too finicky for my taste

pmelanson: as in the MDN's page function? I'd be pretty okay with that. I don't know 100% for sure that `not throwing a RangeError \implies .toLocaleDateString works correctly`, but I'm like 75% sure.

So just to confirm, the `supported()` function from MDN?

pmelanson: thoughts on this? I'll just use the `toLocaleDateStringSupportsLocales()` function from MDN? Or go with the fixed version? I'm def okay with MDN's function, just want to make sure that's what you want.

torhve: @lorenzhs you want to give your thoughts?

pmelanson:
added some work and fixed merge conflicts, but just have to run to class. I'll test one or two things when i get back, don't merge just yet pls

js/glowingbear.js
```diff
@@ -761,6 +761,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
+    $rootScope.supports_formatting_date =
+        function toLocaleDateStringSupportsLocales() {
```
lorenzhs: if you assign it to a variable, the function doesn't need a name:
```js
$rootScope.supports_formatting_date = function() { ... }
```

js/glowingbear.js
```diff
@@ -761,6 +761,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+        // toLocaleDateStringSupportsLocales taken from MDN:
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference
+        // /Global_Objects/Date/toLocaleDateString
+        // #Checking_for_support_for_locales_and_options_arguments
```
lorenzhs: Please don't break lines in URLs, it took me two reads to recognize it. Long lines for long URLs :)

js/handlers.js
```diff
@@ -51,6 +51,7 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
         // Check if the date changed
         if (old_date.valueOf() !== new_date.valueOf()) {
             ++buffer.lastSeen;
+
```
lorenzhs: newline slipped in here

js/handlers.js
```diff
@@ -73,8 +82,21 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
...
+                                                     extra_date_format);
+            } else {
+              // ew ew not more gross code
+              var month_to_word = [
+                "January", "February", "March", "April",
+                "May", "June", "July", "August",
+                "September", "October", "November", "December"];
+              content += month_to_word[new_date.getMonth()] + " " +
+                new_date.getDate().toString();
+              content += (extra_date_format.year === "numeric") ?
```
lorenzhs: this would probably be better expressed as a normal condition, without the ternary operator.

Also, there's no need to make lines *that* short as those above and these

lorenzhs:
Thanks! I added a few more nitpicky comments ;)

js/glowingbear.js
pmelanson: changed.

pmelanson:
Changes made, commits squashed. Tested, although if whoever originally had this could let me know if it's fixed that'd be super.

lorenzhs:
Awesome, thanks so much. It works perfectly on my iPad.

PR #708, opened 2015-12-17
#705 increment buffer.lastSeen on date change message

### LINKED ISSUE #705: Read marker is out of sync due to date change messages
This is a regression introduced by our fancy new date markers, we should fix it.

Filing this issue so we don't forget.

pmelanson:
This seems to fix #705 for me. To test, I clicked on an offending buffer in the bufferlist. There was one date change message, and the read marker was one line too high. I made the change, clicked on the buffer in the hotlist again, and the read marker was now set correctly.


torhve:
Thanks a lot for fixing this!

PR #714, opened 2015-12-19
Implement infinite scrolling

torhve:
I't not perfect but it's a start.

### DIFF
--- a/index.html
+++ b/index.html
@@ -36,6 +36,7 @@
     <script type="text/javascript" src="js/connection.js"></script>
     <script type="text/javascript" src="js/file-change.js"></script>
     <script type="text/javascript" src="js/imgur-drop-directive.js"></script>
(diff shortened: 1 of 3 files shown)


lorenzhs:
I noticed that when switching to a previously not visited buffer (i.e. lines aren't cached) where the initial amount of lines doesn't suffice to fill the screen, it triggers a second fetch. Not sure if intended but pretty cool!

torhve:
Not intentional, as long as the thingy is visible it will happen. But I guess it actually is a good thing?

lorenzhs:
Yeah, I like it, and it only happens once so it won't keep going on forever if there was a giant netsplit and you're using a smart filter.

torhve:
Been running this a few days and I like it. Maybe could merge and see if users finds any problem?

lorenzhs:
yup I agree, it feels pretty nice overall

PR #716, opened 2015-12-20
Chrome WebApp now loads manifest

torhve:


lorenzhs:
so now we have manifest.json, manifest.webapp and webapp.manifest.json?

PR #720, opened 2015-12-26
Add black theme for AMOLED displays, fix theme selector background in dark theme

lorenzhs:
Deprecates / closes #671

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -370,12 +370,12 @@ td.time {
     left: 0;
 }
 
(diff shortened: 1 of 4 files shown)


torhve:
I think it makes sense, I have seen how nice black can be on amoled screens. Wonder if we could detect that in javascript? :-)

lorenzhs:
Given that JS can't even detect whether the screen is a touch screen, I *highly* doubt it :D

PR #721, opened 2015-12-27
Update Angular to 1.4.x series (1.4.8)

lorenzhs:
Angular now has a .$$watchersCount property, which makes our watcher
counting function unnecessary.

Also add sha384 emojione SRI hash


lorenzhs:
I played around with this for a couple of minutes and didn't notice any breakage, everything seems to work fine.

lorenzhs:
I suggest merging this to see if it breaks any corner cases. It's not going to get a lot of testing otherwise.

PR #722, opened 2016-01-02
parse dropbox URLs for image preview plugin (v4)

falaca:
Handles the following scenarios:
-If dl=0 parameter is given, it changes it to dl=1
-If dl parameter is missing, it adds it
-All other query parameters are preserved

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -280,10 +280,25 @@ plugins.factory('userPlugins', function() {
             } else if (url.match(/^http:\/\/(i\.)?imgur\.com\//i)) {
                 // remove protocol specification to load over https if used by g-b
                 url = url.replace(/http:/, "");
-            } else if (url.match(/^https:\/\/www\.dropbox\.com\/s\/[a-z0-9]+\/[^?]+$/i)) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
As Travis said, please use strict equals (`===` and `!==`) and also fix whitespace (use spaces, not tabs; the space between `if` and `(` is missing in line 300). I'd also suggest making this a bit more compact, like so:
```js
            } else if (url.match(/^https:\/\/www\.dropbox\.com\/s\/[a-z0-9]+\//i)) {
                // Dropbox requires a get parameter, dl=1
                var dbox_url = document.createElement("a");
                dbox_url.href = url;
                var base_url = dbox_url.protocol + '//' + dbox_url.host + dbox_url.pathname + '?';
                var dbox_params = dbox_url.search.substring(1).split('&');
                var dl_added = false;
                for (var i = 0; i < dbox_params.length; i++) {
                    if (dbox_params[i].split('=')[0] === "dl") {
                        dbox_params[i] = "dl=1";
                        dl_added = true;
                        // we continue looking at the other parameters in case
                        // it's specified twice or something
                    }
                }
                if (!dl_added) {
                    dbox_params.push("dl=1");
                }
                url = base_url + dbox_params.join('&');
            }
```

lorenzhs:
Thanks!

PR #723, opened 2016-01-02
optimize what gets passed to emojione.unicodeToImage

kurros:
With "non-native Emoji support" enabled, *filters.js* is passing every chunk of text in a buffer to *emojione.unicodeToImage()*. This function is rather CPU expensive.

I profiled switching to the same buffer (that had a half-dozen emoji) multiple times in Google Chrome 47 on my Intel Core i7-4770. GB spent an average of **43.7ms** in *filters.js:161* (*weechat.filter(emojify)*). I added a RegExp.test to check if the chunk includes anything from the Unicode plane that emoji live in. A retest brought the average down to **2.8ms**. Testing on a buffer with no emoji at all didn't even make a blip on the profiler.


lorenzhs:
Thanks, this is a great idea!

Issue #725, opened 2016-01-03
Update "Encryption Instructions" to include SHA256 support


ghostfreeman:
I know this is probably a miniscule nitpick, but there have been some recent developments [[1]](http://arstechnica.com/security/2015/10/sha1-crypto-algorithm-securing-internet-could-break-by-years-end/) regarding the safety of SHA-1 and its now encouraged that new keys be signed as SHA256. The project should take a proactive first step in helping those get onboarded with encrypting their Weechat client by updating to reflect growing security trends. It also appears to be a rather trivial change that I could make a pull request for if need be.

It's a simple argument add to the openssl command listed in "Encryption Instructions" to use SHA256. The actual command would look like this:

```
openssl req -nodes -newkey rsa:4096 -sha256 -keyout relay.pem -x509 -days 365 -out relay.pem -subj "/CN={{IP_ADDRES}}/"
```

torhve:
Yeah! Please submit PR!

ghostfreeman:
PR submitted

PR #726, opened 2016-01-03
#725 Updated wording to include reference to openssl sha256 argument

ghostfreeman:
As discussed in the issue I opened.

Changes are trivial and pass local unit tests.


lorenzhs:
Thanks!

PR #727, opened 2016-01-05
Fixed a circular ref in date change messages

pmelanson:
Before, `line.buffer` was the actual buffer object, which contains the injected date change message itself. Made this just a buffer id, which should be the same as other normal messages.

Not sure if this is exactly how we want it, but in my testing `line.buffer` of an injected date change message and `line.buffer` of a real message are both the same (eg `12b27c0`).

This is probably a good thing to fix. Let me know if you have thoughts.


lorenzhs:
Thanks, good catch!

PR #730, opened 2016-01-07
#705 bugfix datechange injected under read marker

pmelanson:
after #708 was merged, if a date change message is
injected underneath the read marker the read
marker would be one line too low. Now, the read
marker will adjust properly when a date change
message is injected above and below the read
marker.

thanks lorenzhs for spotting that. Does this fix the behavior? I've tested it myself, and it seems to have fixed the problem.

Behavior when:
1. print "old date" to buffer
2. navigate away
3. print "new date" with a different date to the buffer
4. navigate back and take screenshot (see before and after)
before
[before]

after
[after]
 
Yeah, thoughts?

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -41,7 +41,7 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
     };
 
     // inject a fake buffer line for date change if needed
-    var injectDateChangeMessageIfNeeded = function(buffer, old_date, new_date) {
+    var injectDateChangeMessageIfNeeded = function(buffer, manually, old_date, new_date) {
         if (buffer.bufferType === 1) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
It works for the first fetch but when I fetch more lines, the date change markers in those aren't counted against it. That's probably because the read marker is advanced *after* the line is added and the line is added to the array after the read marker was set. Baaaaaaah. I'm not sure what the best way to fix this would be.

pmelanson:
new commit a06f567. Basically, I looked at what normal lines were doing about incrementing buffer.lastSeen and replicated it. Should've done this earlier, really, but I'm doing it now. In my testing, this gives the correct behaviour when the read marker is below and above a date change message, as well as when I fetch new lines. Works even when the injected date change message is generated by handlers.js:L334 .

Thoughts?

PR #735, opened 2016-01-16
Include version information in settings dialog, update version numbers

### LINKED ISSUE #734: Include version number in releases
I would love to be able to check what release I cloned into my web servers directory from within the GB interface. This way when I check the project I would be able to reference the GB version running on my server without having to ssh into the web host's directory and run a `git status`.

Perhaps you could place a version for major releases somewhere in the Setttings frame. Presently I think a nice place for it would be inside to the left of the close button, below the dividing line on the left-hand side. It could say something like:
* `Glowing Bear version 0.5.0`.

Beneath could be a link to the releases and text that could say something like:
* Check for [new releases](https://github.com/glowing-bear/glowing-bear/releases).
With a link just like the one above. I think that would be a great feature to include and would take minimal effort. :smiley:

lorenzhs:
Closes #734

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -15,6 +15,10 @@ a {
     cursor: pointer;
 }
 
+.version {
+   margin-right: 1em;
+}
+
 .hidden-bracket {
   position: absolute;
   left: -1000px;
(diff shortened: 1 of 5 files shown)


torhve:
Is this documented in our release document? I.E how to make releases => remember to upgrade numbers

Maffblaster:
Glad to see you work on this so soon. :)

lorenzhs:
@torhve it is now: https://github.com/glowing-bear/glowing-bear/wiki/Releasing

PR #739, opened 2016-01-26
Enhancements for a hypothetical Windows app

kddlb:
*original title: Enable the use of UWP (Windows 10) toasts*

Obviously this would work only if Glowing Bear is running inside a Windows Web App container.

### DIFF
--- a/js/notifications.js
+++ b/js/notifications.js
@@ -45,7 +45,28 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
                     tag: 'gb-highlight-vib'
                 });
             });
+        } else if (typeof Windows !== 'undefined' && typeof Windows.UI !== 'undefined' && typeof Windows.UI.Notifications !== 'undefined') {
+
+            var winNotifications = Windows.UI.Notifications;
+            var toastNotifier = winNotifications.ToastNotificationManager.createToastNotifier();
+            var template = winNotifications.ToastTemplateType.toastText02;
+            var toastXml = winNotifications.ToastNotificationManager.getTemplateContent(template);
+            var toastTextElements = toastXml.getElementsByTagName("text");
+
+            toastTextElements[0].appendChild(toastXml.createTextNode(title));
+            toastTextElements[1].appendChild(toastXml.createTextNode(body));
+
+            var toast = new winNotifications.ToastNotification(toastXml);
+
+            toast.onactivated = function() {
+                models.setActiveBuffer(buffer.id);
+                window.focus();
+            };
+
+            toastNotifier.show(toast);
+
         } else {
+
             var notification = new Notification(title, {
                 body: body,
                 icon: 'assets/img/favicon.png'
@@ -74,7 +95,9 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
             notification.onclose = function() {
                 delete notifications[this.id];
             };
+
         }
+
     };
 
 


kddlb:
By the way, a hypothetical Windows Store app would contain no code, just a pointer to https://latest.glowing-bear.org/

js/notifications.js
```diff
@@ -46,34 +46,65 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
...
-            // Save notification, so we can close all outstanding ones when disconnecting
-            notification.id = notifications.length;
-            notifications.push(notification);
+            
+            //If it detects the Windows object, that is, running on a Hosted Web App container, use the UWP toasts
+            if (typeof Windows !== 'undefined' && 
+                typeof Windows.UI !== 'undefined' && 
+                typeof Windows.UI.Notifications !== 'undefined') {
+                    
+                    var winNotifications = Windows.UI.Notifications;
```
lorenzhs: Can you fix the indentation here? It shouldn't be two blocks. AIso I find the blank lines between the declarations make this unnecessarily hard to read

js/notifications.js
```diff
@@ -46,34 +46,65 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
...
+                    var toast = new winNotifications.ToastNotification(toastXml);
+
+                    toast.onactivated = function() {
+                        models.setActiveBuffer(buffer.id);
+                        window.focus();
+                    };
+                    
+                    toastNotifier.show(toast);
+                    
+                } else {
```
lorenzhs: indentation is wrong because of the above, should be one block less

js/notifications.js
```diff
@@ -46,34 +46,65 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
...
-                body: body,
-                icon: 'assets/img/favicon.png'
-            });
-
-            // Save notification, so we can close all outstanding ones when disconnecting
-            notification.id = notifications.length;
-            notifications.push(notification);
+            
+            //If it detects the Windows object, that is, running on a Hosted Web App container, use the UWP toasts
+            if (typeof Windows !== 'undefined' &&
```
lorenzhs: please make this an else if with the above condition, that would preserve the old code as-is without indentation changes which make code changes unnecessarily hard to trace

lorenzhs:
I don't have a Windows system to test, but this looks like a cool thing to have. Could you just fix the indentation mess and squash the commits, please?  :+1:

kddlb:
By the way, this is how the app would run:

https://microsoftedge.github.io/WebAppsDocs/en-US/win10/CreateHWA.htm

The only pet-peeve is that you have to whitelist the inline (preview) and script domains:

``` xml
<uap:ApplicationContentUriRules>
    <uap:Rule Match="https://*.imgur.com" Type="include" WindowsRuntimeAccess="none" />
    <uap:Rule Match="https://cdnjs.cloudflare.com" Type="include" WindowsRuntimeAccess="none" />
    <uap:Rule Match="https://*.dailymotion.com" Type="include" WindowsRuntimeAccess="none" />
    <uap:Rule Match="https://*.youtube.com" Type="include" WindowsRuntimeAccess="none" />
    <uap:Rule Match="http://*.allocine.fr" Type="include" WindowsRuntimeAccess="none" />
</uap:ApplicationContentUriRules>
```

js/notifications.js
```diff
@@ -54,7 +75,7 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
             // Save notification, so we can close all outstanding ones when disconnecting
             notification.id = notifications.length;
             notifications.push(notification);
-
+
```
lorenzhs: Whoop whoop, whitespace police!

lorenzhs:
Hm, the whitelisting requirement is indeed problematic, as glowing bear can embed images and videos from arbitrary sites by matching the file extension. We should probably detect  running in the Windows app and not offer those embeds to avoid frustration.

PR #746, opened 2016-02-19
Add preview of .mp3, .ogg and .wav links via <audio> tag.

### LINKED ISSUE #745: Add <audio> embeds for ogg, mp3
URLs to audio files don't have embeds at the moment, while images and video do. We should fix that oversight.

simeng:
This fixes #745


torhve:
SOUNDS good to me! Thanks

PR #751, opened 2016-03-10
Fix a typo issue in index.html

remimarenco:
Just a typo issue when having Encryption error ;)


torhve:
Thank you for fixing this!

PR #754, opened 2016-03-18
Changed word-break property for td.message to normal.

RMoDOOM:
The messages break-between-any-letter styling was bugging me, as it made some messages very difficult to read, so I created a stylish script with these rules. It seems to be working properly with that, and isn't interfering with links, so I added it directly to the stylesheet here.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -38,18 +38,16 @@ td.message {
     width: 100%;
     padding: 1px 1px 1px 5px;
 
--ms-word-break: break-all;
-    word-break: break-all;
+    -ms-word-break: normal;
+    word-break: normal;
 
-    /* Non standard for webkit */
-    word-break: break-word;
-
--webkit-hyphens: auto;
-   -moz-hyphens: auto;
+    -webkit-hyphens: auto;
+    -moz-hyphens: auto;
     -ms-hyphens: auto;
         hyphens: auto;
 
 }
+
 
 #readmarker {
     margin-top: 5px;


torhve:
Reason we have it like this is... complicated, because... browsers!
Have you tested thoroughly in all major browser engines that this works as intended?

RMoDOOM:
I didn't expect such a quick response, was in the middle of doing so (probably jumped the gun on the PR oops). Firefox 44.0.2 works and IE11 works. Chrome won't accept my SSL certificate because oh no self-signed, but I should be able to test that (and MS Edge) later; I don't have a non-SSL relay running and I'm currently on Win7.

lorenzhs:
You can test chrome by going to the relay URL (https://weechathost:relayport/weechat) and accepting the cert for the current session.

The main issue (and reason for the current value) was that for *some* long lines, there would be horizontal scroll bars. It only happened sometimes, definitely not for each long line, so it's a bit awful to test. Best try various different long messages (including long URLs and such) and resizing the browser window and see if scroll bars appear at any width.

torhve:
Make sure you check with external content, like youtube, images, etc. Text is not enough. And if you could put some screenshots aswell would be nice to illstrutate how this behaves

lorenzhs:
Any updates?

RMoDOOM:
Sorry actual job & life got in the way of this. I've been using it on both chrome and firefox for a few weeks without noticing any oddities or problems.

Also noticed that there's a custom CSS field, and I've been using that instead. Maybe a note on the front page (or near that box) with some sample CSS, including this thing, would be sufficient?

If I don't have time to more thoroughly test it in the next week or so, I'll withdraw this PR.

lorenzhs:
We have a wiki article with some CSS snippets, we probably should link to it in a more prominent place. https://github.com/glowing-bear/glowing-bear/wiki/CSS-Customizations

RMoDOOM:
I don't have the time to thoroughly test this. Apologies.

Issue #598, opened 2015-04-15
Away status


SamuelMoraesF:
Will be great an box(maybe at the right of the input bar) with the user's nickname and the away status.

![captura de tela de 2015-04-14 22 38 51](https://cloud.githubusercontent.com/assets/3705263/7150847/3a334cf8-e2f7-11e4-931a-53102b20f7f9.png)

torhve:
If someone wants to submit a patch for that maybe we could include it.

lorenzhs:
If we want to add this, it would be important not to show it on mobile, as space in the input bar is scarce.

t3chguy:
Maybe on Mobile just show a small - single character symbol instead of (away)

lorenzhs:
That would be unintuitive as it has absolutely no meaning if you don't already know what it means. UI is hard :(

Maffblaster:
I like the idea of an away status for the web UI (as displayed in @SamuelMoraesF's image above).

Bakke:
How about putting nickname and status in the top-left corner, with the Glowing Bear-logo and unread count?

lorenzhs:
yeah that might work if it doesn't look to detached from the rest of the UI

Techman-:
>That would be unintuitive as it has absolutely no meaning if you don't already know what it means. UI is hard :(

What you could do is make the "A" the same color as away nicks in weechat.

PR #757, opened 2016-03-23
Add a button for leaving a channel

doanac:
This adds a button to close the active channel.

Signed-off-by: Andy Doan <andy.doan@linaro.org>

### DIFF
--- a/index.html
+++ b/index.html
@@ -244,6 +244,9 @@ $ openssl req -nodes -newkey rsa:4096 -keyout relay.pem -x509 -days 365 -out rel
         </div>
 
         <div class="actions pull-right vertical-line-left">
+          <a ng-click="leave()" ng-if="activeBuffer().type === 'channel' || activeBuffer().type === 'private'" title="Leave {{ activeBuffer().trimmedName || activeBuffer().fullName }}" href="#">
+            <i class="glyphicon glyphicon-remove"></i>
+          </a>
           <a class="settings-toggle" ng-click="showModal('settingsModal')" title="Options menu" href="#">
             <i class="glyphicon glyphicon-cog"></i>
           </a>
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -612,6 +612,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         $scope.connectbuttonicon = 'glyphicon-chevron-right';
         connection.disconnect();
     };
+    $scope.leave = function() {
+        connection.sendMessage("/leave");
+    };
     $scope.reconnect = function() {
         var bufferId = models.getActiveBuffer().id;
         connection.attemptReconnect(bufferId, 3000);


lorenzhs:
@torhve @cormier thoughts?

torhve:
Only conflicting thouhgs.

Nice to have, but what other buttons should we then also add? Maybe a context menu somewhere would make more sense, where we could add lots of buttons that aren't visible all the time?

I fear clutter, but I agree that our interface is not nice to use with touch/mouse.

lorenzhs:
Context menu is no good on mobile, but I have the same conflicting thoughts you do.

Bakke:
Dropdown menu?

doanac:
A context menu would be great for desktop. It seems more intuitive to
right-click and channel and select "leave". However, this type change would
be far too invasive for me to offer up as a first commit to this project. I
was just hoping this idea might motivate someone to find a nice alternative.

cormier:
This feature is interesting. However I don't think that we should have buttons other than settings and exit directly in the title bar. A context menu or a dropdown seem to be the best way to provide secondary action. 

@lorenzhs : is the issue of the context menu on mobile an issue of triggering the context menu? Could a context menu be interesting on mobile if was triggered by a long press ? Libraries like hammerjs seem to provide an interesting way to provide long press support.

If the context menu is out of question, I would go for a dropdown. However I feel like dropdowns are more suitable for choosing values in a form than triggering actions. Maybe it's only me :-)

lorenzhs:
hmm, not sure I like the ideas of context menu on mobile. It's "hidden UI", magical features that you don't know exist until you trigger them by accident. Not sure what would work best.

torhve:
Proposal:
 * Make a new "bar"  at the bottom left of the button list.
 * Move global buttons into this bar.
 * Have leave button, and nicklist button on the top right. Maybe at max 2-3 buttons.
 * Put current nick and away status also in the bottom left bar.

Quick and dirty mspaint example:

glowing 20bear_2016-04-11_16-06-34 1

lorenzhs:
I think having the global buttons at the bottom left is counterintuitiv. They won't be visible on mobile unless you open the bufferlist. If anything, I'd support the reverse (add buffer-related buttons to the bottom of the bufferlist)

torhve:
I looked at a few other apps, and some do it the way I propose. But it's super hard to design a good interface, so I'm just throwing this out as a starting point for further discussion.

And not having global actions visible all the time sounds like an improvement. You still have to swipe to access channel switching, and that's a far more used action than Settings/Disconnect.

Bakke:
I think @torhve is on to something. I would suggest putting "Settings" and "Disconnect" in the top-left corner. All global stuff could go here, including nick and away-status from issue #598. Channel actions, like leave and nicklist-button, should be in the top-right corner.

Vaesper:
Interesting thoughts in this thread. Would a "join channel" button conceivably have a nice place in this as well?

I personally think that a "close buffer" kind of button would not be misplaced under a "buffer options" kind of button (where perhaps you could e.g. also set the notify level for the buffer?). I don't know how others use IRC but my channel list is _mostly_ stable so I don't need this "close buffer" button in plain view (and prone to accidental pushing when fat fingering), but a way to do it without typing commands would definitely be nice (after all, if we wanted a command line we wouldn't be using GB).

AStove:
Would an action button at the bottom left be an option? The action button then opens a dialog. The dialog can be the default system dialog on mobile.

- Isn't a hidden feature
- Room for more future options
- Minimizes clutter (no extra elements on mobile since the bar is normally closed)
- Behavior can depend on what's selected, if nothing is selected join could be used.
- On mobile would have to inhibit default actions for a longtouch so channels can be selected without the sidebar closing.

image

PR #761, opened 2016-03-30
WebApp not working: Need to remove /glowing-bear prefix from urls.

magnific0:
The WebApp installation through firefox wasn't working. Turns out the urls specified in manifest.webapp point all ponit to resources inside the glowing-bear directory, which doesn't exist. 

The only error when launching the webapp is

> Error response
> 
> Error code: 404
> 
> Message: File not found.
> 
> Error code explanation: HTTPStatus.NOT_FOUND - Nothing matches the given URI.

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -1,12 +1,12 @@
 {
   "name": "Glowing Bear",
   "description": "WeeChat Web frontend",
-  "launch_path": "/glowing-bear/index.html",
+  "launch_path": "/index.html",
   "icons": {
-    "128": "/glowing-bear/assets/img/glowing_bear_128x128.png",
-    "60": "/glowing-bear/assets/img/glowing_bear_60x60.png",
-    "90": "/glowing-bear/assets/img/glowing_bear_90x90.png",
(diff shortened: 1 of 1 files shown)


lorenzhs:
See #606, this is more complicated than it looks.

lorenzhs:
I think now that we have TLS on glowing-bear.org we may actually be able to drop the prefix though. In any case, Mozilla is removing web apps from Firefox so it won't matter for long. @torhve

torhve:
Yes, we will remove this firefox install functionality, since Mozilla is removing it soon.

lorenzhs:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1238079 - gone with firefox 47

lorenzhs:
@torhve how shall we proceed? I think we should merge this even if the functionality gets removed by mozilla as the majority of use cases where someone would install the firefox thingy doesn't have `/glowing-bear/`

lorenzhs:
Closing this now because Mozilla made it obsolete.

PR #762, opened 2016-03-30
Rudimentary Electron (Atom-Shell) support

torhve:
This patch serves as a starting point for making electron based Glowing
Bear builds. Primary motivation is that Mozilla is removing support for
their web runtime, so then Electron is a new way to get standalone
application running outside of browser.

This adds 2 new files that will be the main html and main js for
electron. Those files simply load the rest of the app.

It also adds two new npm run commands to build windows and darwin builds
for electron. Darwin is untested so far.

Bulding packages requires npm package `electron-packager`

There's still some problems with notifications.
And badges for the dock in Darwin is not implemented.
We could also put all keyboard shorcuts in menus.

Thanks to @lukas2511 for some guidance on the wrappers and commands.

### DIFF
--- a/electron-main.js
+++ b/electron-main.js
@@ -1,0 +1,198 @@
+(function() {
+    'use strict';
+    const electron = require('electron');
+    const app = electron.app;  // Module to control application life.
+    const BrowserWindow = electron.BrowserWindow;  // Module to create native browser window.
+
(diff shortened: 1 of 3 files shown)


torhve:
Closing this as Bakke has a new coming.

PR #764, opened 2016-03-30
Electron (Atom-Shell) support

torhve:
This patch serves as a starting point for making electron based Glowing
Bear builds. Primary motivation is that Mozilla is removing support for
their web runtime, so then Electron is a new way to get standalone
application running outside of browser.

This adds a couple new files that will only be useful in electron builds and not interfere with the other usages of the project.

It also adds two new npm run commands to build windows, darwin and linux releases
for electron.

Bulding packages requires npm package `electron-packager` which has been added to package.json even if it's not used for all types of gb-usage. A simple makefile has been added to the project to help maintain different build/release commands, like pulling down 3rdparty resources locally and actually produce executables.

There's still some problems with focus on win32, after alt-tabing.
And badges for the dock in Darwin is implemented.
Setoverlayicon for win32 is implemented, but not using a cool notification icon, yet. (Can be improved later)
We could also put all keyboard shorcuts in menus.

Thanks to @lukas2511 for some guidance on the wrappers and commands.


lorenzhs:
Some documentation and a linux build would be nice!

lorenzhs:
I ran `electron-packager  . "Glowing Bear" --platform=linux --arch=x64 --version=0.37.3 --overwrite --version-string.FileDescription="Glowing Bear"` which worked fine, I guess we should add that as a linux target. But emoji don't work!

lorenzhs:
Opening the inspector also opens it both in a new window as well as in the same window on the side. Wat.

lorenzhs:
You should also exclude the "node_modules" folder from the build. I guess we can keep "bower_components" and switch to using those instead of the CDN for a truly local build. That should probably fix emoji as well.

electron.makefile
```diff
@@ -13,8 +13,8 @@ copylocal:
 
 # modify index.html to use local files
 uselocal: copylocal
-	sed -i 's,https://cdnjs.cloudflare.com/ajax/libs/[^\"]*/,3rdparty/,g' index.html
-	sed -i 's, integrity=\".*\" crossorigin=\"anonymous\",,' index.html
+	sed -i .bak 's,https://cdnjs.cloudflare.com/ajax/libs/[^\"]*/,3rdparty/,g' index.html
```
lorenzhs: I get `sed: -e expression #1, char 1: unknown command: `.'` on Linux now

torhve:
@cormier would you mind reviewing this rather large changeset? Since both me and @lorenzhs has worked on it, we might not be the best candidates to self-review.

20after4:
I've been working on something similar, only tested on linux. I am currently using it for my own glowing bear instance but I don't think it's quite ready for upstreaming. There may be something useful though: https://github.com/20after4/electric-gb/

20after4:
Also: lgtm

lorenzhs:
@cormier what do you think?

@20after4 cool - haven't had a thorough look at it yet but is there anything you miss in our branch?

20after4:
Mine has a tray icon which mirrors the favicon with highlight / unread count. I'm sure I could submit a pull request to add that to your branch. Mine is more messy because I was learning electron and experimenting with a lot of things not really intending to make something ready for public consumption.

Good work on this by the way, it looks good to me and I'll gladly work on cleaning up some of my code and submit another PR after this lands.

lorenzhs:
Does anyone have an issue with merging this branch? If not I'll just click merge on the weekend.

lorenzhs:
:shipit:

torhve:
:+1:

Techman-:
I guess I'll add my feedback here in addition to the IRC channel.

* CTRL + F would be used to toggle full screen. I use it to search text on the page. If Electron doesn't have this feature then I guess it doesn't matter then.
* F12 is also used to toggle dev tools, so perhaps that keyboard shortcut should be hooked as well.

lorenzhs:
@Techman- the way I see it it's only Ctrl + Cmd + F that's bound to fullscreen (on Mac), and F11 everywhere else?

Also the electron app isn't the best medium for debugging anyway so pressing Ctrl+Shift+I instead of F12 shouldn't be a problem imho

PR #765, opened 2016-03-31
Imgurhttps

kurros:
When using the Imgur upload tool (which uses HTTPS), the API currently responds with an HTTP URL. We should share it as an HTTPS URL.


torhve:
I like this!

Would you mind rebasing this to remove the Merge-commit?

kurros:
Think I figured it out!

lorenzhs:
Awesome, thanks!

PR #766, opened 2016-04-03
Switch math rendering to KaTeX

### LINKED ISSUE #760: Replace MathJax with KaTeX or something
MathJax is impossible to use with a CSP that deserves the name, it's also un-bundleable in the app. We should use something else.

This blocks #437

lorenzhs:
It's faster and easier to package. This way we'll be able to have math in electron and cordova without fetching the remote scripts. We'll also be able to use a CSP!

Fixes #760

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -175,15 +175,23 @@ weechat.filter('emojify', function() {
     };
 });
 
-weechat.filter('mathjax', function() {
+weechat.filter('latexmath', function() {
     return function(text, selector, enabled) {
-        if (!enabled || typeof(MathJax) === "undefined") {
+        if (!enabled || typeof(katex) === "undefined") {
             return text;
         }
         if (text.indexOf("$$") != -1 || text.indexOf("\\[") != -1 || text.indexOf("\\(") != -1) {
-            // contains math
-            var math = document.querySelector(selector);
-            MathJax.Hub.Queue(["Typeset",MathJax.Hub,math]);
+            // contains math -> delayed rendering
+            setTimeout(function() {
+                var math = document.querySelector(selector);
+                renderMathInElement(math, {
+                    delimiters: [
+                        {left: "$$", right: "$$", display: false},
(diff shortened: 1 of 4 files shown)


js/filters.js
```diff
@@ -175,15 +175,23 @@ weechat.filter('emojify', function() {
...
         }
         if (text.indexOf("$$") != -1 || text.indexOf("\\[") != -1 || text.indexOf("\\(") != -1) {
-            // contains math
-            var math = document.querySelector(selector);
-            MathJax.Hub.Queue(["Typeset",MathJax.Hub,math]);
+            // contains math -> delayed rendering
+            setTimeout(function() {
+                var math = document.querySelector(selector);
+                renderMathInElement(math, {
+                    delimiters: [
```
lorenzhs: should I factor that out somewhere?

js/glowingbear.js
```diff
@@ -381,27 +381,37 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     });
 
+    // Inject a javascript (KaTeX)
```
torhve: Wrong comment?

js/glowingbear.js
```diff
@@ -381,27 +381,37 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+    var inject_script = function(script_url) {
+        var script = document.createElement("script");
+        script.type = "text/javascript";
+        script.src  = script_url;
+        var head = document.getElementsByTagName("head")[0];
+        head.appendChild(script);
+    };
+    // Inject a stylesheet (KaTeX or theme)
+    var inject_css = function(css_url, id) {
```
torhve: Maybe move both of these functions to utils.js?

js/glowingbear.js
lorenzhs: why? it injects a javascript and is used by the KaTeX thing

PR #768, opened 2016-04-08
Make the settings close button more visible for dark themes

tnguyen14:
Before: 
screen shot 2016-04-08 at 4 54 18 pm

After:
screen shot 2016-04-08 at 4 54 03 pm


lorenzhs:
Good idea, thanks!

PR #769, opened 2016-04-08
more readable messages with increased line-height

tnguyen14:
Increasing the message `line-height` makes them a bit more readable for me.

I could foresee however that someone might prefer the current more compact setting of the messages. If that is so, I could create a setting that will switch the `line-height`, so the increased line-height could be by default or an opt-in setting.

### DIFF
--- a/css/themes/dark-spacious.css
+++ b/css/themes/dark-spacious.css
@@ -1,0 +1,5 @@
+@import "dark.css";
+
+td.message {
+    line-height: 1.4;
+}
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -33,7 +33,6 @@ td.prefix {
     border-right: 1px solid #444;
 }
 td.message {
-    line-height: 1.4;
     overflow: hidden;
     vertical-align: top;
     width: 100%;
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -33,6 +33,7 @@ td.prefix {
     border-right: 1px solid #444;
 }
 td.message {
+    line-height: 1.4;
     overflow: hidden;
     vertical-align: top;
     width: 100%;


torhve:
I like our compact look in the default theme. 

I feel like changes like this is exactly why we support custom css, for those minute changes that personally feels nice.

But I would like for someone to create a less compact theme with more space all around.

tnguyen14:
Yeah that makes sense.

I do have this style in my custom CSS. The reason I thought it would make sense to make it a configurable setting is when I look at some other chat tools like Slack or HipChat, most of them have the option to display more readable messages vs. a compact version.

torhve:
We too have that option, but in a better form, fully replacable themes!

It's just that we haven't gotten many submissions for themes yet :-)

tnguyen14:
If you think that this specific feature should be a theme, I could spin it off into one. Then perhaps others could chime in on where more spaces should be added. Right now, the only place that seems urgent to me is the messages.

If I was to do that, what theme should I base it off of? dark, black or light? Or should there be a less compact version for all 3?

lorenzhs:
I think dark would be the best starting point, as it's our default theme and the most "polished" one (if you can say such a thing about our UI). The light theme needs work anyway, and black is only a dozen line change over dark, see https://github.com/glowing-bear/glowing-bear/blob/master/css/themes/black.css, so the changes to dark would carry over to black directly.

tnguyen14:
@lorenzhs That makes sense to me. So should it be a new theme that extends `dark`, just like the black one, or should these changes be included directly into the `dark` theme?

lorenzhs:
We'd like to keep the existing dark theme, I think a "spacious" theme would best be implemented in the same way as "black", by including the dark theme and building upon it. Does that sound workable?

tnguyen14:
I've added a new `dark-spacious` theme to implement this change.

Any suggestions to other places where we can make it more "spacious" is welcome.

css/themes/dark-spacious.css
```diff
@@ -0,0 +1,5 @@
+import "dark.css";
```
lorenzhs: the correct syntax is `@import`, not `import`

tnguyen14:
@lorenzhs thanks for pointing that out. my bad. I've fixed it.

vrutkovs:
This makes nicks having the same line-height

Using ```tr.bufferline``` instead makes it much better

lorenzhs:
hmm good idea, would you like to send a PR?

tnguyen14:
@vrutkovs why would nicks having the same line-height bad?

vrutkovs:
My fault, I meant to say that nick don't get 1.4 line-height

It gets displayed with a small vertical margin here:
selection_030

Issue #741, opened 2016-01-26
Fix issue where many connections are opened


lorenzhs:
@weechatter had a mysterious issue where Glowing Bear would rack up loads of connections in a loop. From his logs (Android app at https://github.com/glowing-bear/glowing-bear/tree/df23985cdacd3c2ba748410ff2c526f26a1b6c2d)

```
Failed to load resource: net::ERR_FILE_NOT_FOUND
angular.min.js:102 Service Worker is supported
angular.min.js:102 Connecting to URL:  ws://foo.bar:1234/weechat
angular.min.js:102 Service Worker err: DOMException: Failed to register a ServiceWorker: The URL protocol of the current origin ('file://') is not supported.
angular.min.js:102 Connected to relay
2angular.min.js:102 Fetching  42  lines
http://fooo.bar/baz.png Failed to load resource: the server responded with a status of 404 (Not Found)
2http://pbs.twimg.com/media/foobar.jpg Failed to load resource: the server responded with a status of 404 (Not Found)
angular.min.js:102 Disconnecting from relay
angular.min.js:102 Disconnected from relay
angular.min.js:102 Service Worker is supported
angular.min.js:102 Connecting to URL:  ws://foo.bar:1234/weechat
angular.min.js:102 Service Worker err: DOMException: Failed to register a ServiceWorker: The URL protocol of the current origin ('file://') is not supported.
angular.min.js:102 Connected to relay
angular.min.js:102 Fetching  42  lines
2angular.min.js:102 Fetching  Infinity  lines
angular.min.js:102 Fetching  35  lines
angular.min.js:102 Unhandled event received: _buffer_cleared
4angular.min.js:102 Fetching  35  lines
angular.min.js:102 Fetching  61  lines
angular.min.js:102 Fetching  51  lines
angular.min.js:102 Fetching  32  lines
8angular.min.js:102 Unhandled event received: _buffer_cleared
4angular.min.js:102 ReferenceError: emojione is not defined
    at m.$scope.inputChanged (inputbar.js:29)
    at angular.min.js:200
    at m.$eval (angular.min.js:126)
    at angular.min.js:216
    at angular.min.js:228
    at q (angular.min.js:7)
    at $$writeModelToScope (angular.min.js:228)
    at angular.min.js:227
    at h (angular.min.js:226)
[...]

Issue #740, opened 2016-01-26
Need a lock on the connection instance


lorenzhs:
We need to make sure that at no point more than one connection can be open. That this isn't currently the case sometimes really messes up things, e.g. #610 or the issue @weechatter described

PR #770, opened 2016-04-10
[PLS TEST] Add a lock on the connection

### LINKED ISSUE #610: Reconnect sometimes results in duplicated line data
On occassion, glowing bear shows duplicate data. Android weechat app and tmux session of weechat don't show this. Screenshot attached. This is for comments, status messages, join/parts, etc...
screen shot 2015-05-13 at 11 29 17 am

lorenzhs:
This should fix #610 and #740 (probably also #741) but needs more testing

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -15,6 +15,9 @@ weechat.factory('connection',
     var connectionData = [];
     var reconnectTimer;
 
+    // Global connection lock to prevent multiple connections from being opened
+    var locked = false;
+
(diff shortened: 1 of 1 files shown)


torhve:
Been testing this, seems to work fine. Cannot reproudce any dupes or problems with connection with this patch.

Will merge for now, and we can see if it triggers problems, or if it actually fixes dupelines or not!

lorenzhs:
Awesome, thanks!

PR #773, opened 2016-04-21
Add a pastebin plugin

pelmers:
I see a pastebin posted today and wonder, "why doesn't glowing-bear give me preview?"
Ah, I find there is no pastebin plugin.

Hope this is useful.


lorenzhs:
neat, thanks!

PR #780, opened 2016-05-20
Update Bootstrap to 3.3.6

vrutkovs:
Supersedes #538, it seems accordion issues have been fixed


lorenzhs:
Is there a reason you're switching from Cloudflare's CDN to Bootstrap's? I'd rather we use Cloudflare for everything.

vrutkovs:
Agreed, switched back to Cloudflare

torhve:
Thanks!

Issue #782, opened 2016-05-29
Display buffer timestamps in a matter consistent with detected locale


Techman-:
(This is a feature request)
I live in the United States, so following strftime my optimal timestamp in this case would be formatted as `%I:%M:%S %p`. [This site](http://foragoodstrftime.com/) is a really neat cheat-sheet. It ends up looking like `08:49:47 PM`. Even though I used strftime for this, it can be converted to what AngularJS uses by [their docs](https://docs.angularjs.org/api/ng/filter/date).

Slightly off track here, but I wonder why [a strftime library](https://www.google.com/search?q=strftime+javascript) for glowing-bear wasn't used.

torhve:
Angular provides functions for formatting time, that's what we use: https://docs.angularjs.org/api/ng/filter/date

torhve:
For those looking to implement this, take a look at the function `injectDateChangeMessageIfNeeded` in file `js/handlers.js` that already does some locale stuff.

pmelanson:
@Techman- made a PR, from my testing it now puts 12h time format if the system locale is 12h time. Do you want to test to make sure? https://latest.glowing-bear.org/pull/783/

torhve:
GB will now use weechat configuration, so this should work now.

Techman-:
Is this working on the main website now?

lorenzhs:
https://latest.glowing-bear.org/ for now

PR #783, opened 2016-06-07
Bufferline timestamps now try to detect locale

pmelanson:
Follows the same pattern as `injectDateChangeMessageIfNeeded` in file `js/handlers.js`, roughly.

Adds a bit more complexity to figure out what the user's locale time string is.

Also, moves a little bit of code complexity from index.html to js/models.js . The reasoning behind this is the format of the time string will only be figured out when javascript code is being executed, and if we want to apply a different style to the time separators as we do now, then we'll have to do that in the JS code.

Prototype PR, so for sure let me know if there's anything to change. For example, [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#Performance) suggests I can use `Intl.DateTimeFormat` something or other to get better performance for large amounts of dates.

Thoughts?

### DIFF
--- a/index.html
+++ b/index.html
@@ -289,7 +289,7 @@ $ openssl req -nodes -newkey rsa:4096 -keyout relay.pem -x509 -days 365 -out rel
             <tr class="bufferline">
               <td class="time">
                 <span class="date" ng-class="::{'repeated-time': bufferline.shortTime==bufferlines[$index-1].shortTime}">
-                  <span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'HH')"></span><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'mm')"></span><span class="seconds"><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'ss')"></span></span>
+                  <span class="cof-chat_time cob-chat_time coa-chat_time timestamp-no-seconds" ng-bind-html="::bufferline.formattedTime"></span>
                 </span>
               </td>
               <td class="prefix"><a ng-click="addMention(bufferline.prefix)"><span class="hidden-bracket">&lt;</span><span ng-repeat="part in ::bufferline.prefix" ng-class="::part.classes" ng-bind="::part.text|prefixlimit:25"></span><span class="hidden-bracket">&gt;</span></a></td><!--
--- a/js/models.js
+++ b/js/models.js
@@ -347,6 +347,24 @@ models.service('models', ['$rootScope', '$filter', function($rootScope, $filter)
         var buffer = message.buffer;
         var date = message.date;
         var shortTime = $filter('date')(date, 'HH:mm');
+        var formattedTime = "";
+
+        if ($rootScope.supports_formatting_date) {
+          formattedTime = date.toLocaleTimeString([], {hour: "2-digit",
+                                                 minute: "2-digit",
+                                                 second: "2-digit"});
+        } else {
+          formattedTime = $filter('date')(date, 'HH:mm:ss');
+        }
+
+        formattedTime = formattedTime.replace(/ /g, "&nbsp;");
+        // If in 12h time the hour field has only one number, zero pad it
+        formattedTime = formattedTime.replace(/^(\d)(:|\.)/, '0$1$2');
+        // Wrap the first time separator in a span
+        formattedTime = formattedTime.replace(/(:|\.)/, '<span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">$1</span>');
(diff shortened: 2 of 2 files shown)


css/glowingbear.css
```diff
@@ -326,6 +326,7 @@ tr.bufferline {
 td.time {
     padding: 1px 5px 1px 1px;
     vertical-align: top;
+    text-align: right;
```
pmelanson: having a 12-hour time locale stops the hour field returned from `Date.toLocaleTimeString()` from being zero-padded, so if we have something like:

```
9:59 pm [text]
10:00 pm [text]
```

This change makes the above turn into:


```
 9:59 pm [text]
10:00 pm [text]
```

pmelanson:
Fixes #782

pmelanson:
oh man totally forgot to keep support for the seconds option, I'll get that tomorrow X/

index.html
```diff
@@ -289,7 +289,7 @@ <h4 class="panel-title">
             <tr class="bufferline">
               <td class="time">
                 <span class="date" ng-class="::{'repeated-time': bufferline.shortTime==bufferlines[$index-1].shortTime}">
-                  <span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'HH')"></span><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'mm')"></span><span class="seconds"><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'ss')"></span></span>
+                  <span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind-html="bufferline.formattedTime"></span>
```
lorenzhs: Can you bind this once with the "::" syntax? Reduces watchers and improves performance

pmelanson:
addressed. lmk if my angular flailings are wrong again.

index.html
```diff
@@ -288,8 +288,9 @@ <h4 class="panel-title">
...
               <td class="time">
-                <span class="date" ng-class="::{'repeated-time': bufferline.shortTime==bufferlines[$index-1].shortTime}">
-                  <span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'HH')"></span><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'mm')"></span><span class="seconds"><span class="cof-chat_time_delimiters cob-chat_time_delimiters coa-chat_time_delimiters">:</span><span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind="::(bufferline.date|date:'ss')"></span></span>
+                <span ng-if="settings.showtimestamp" class="date" ng-class="::{'repeated-time': bufferline.shortTime==bufferlines[$index-1].shortTime}">
+                  <span ng-if="!settings.showtimestampSeconds" class="cof-chat_time cob-chat_time coa-chat_time" ng-bind-html="::bufferline.formattedTime"></span>
+                  <span ng-if="settings.showtimestampSeconds" class="cof-chat_time cob-chat_time coa-chat_time" ng-bind-html="::bufferline.formattedTimeSeconds"></span>
```
lorenzhs: These `ng-if`s introduce 2 watchers per line - I suggest just refreshing the buffer when that setting is changed (as a setting change hook) and using the single-binding syntax (`::`) here as well

index.html
```diff
@@ -276,7 +276,7 @@ <h4 class="panel-title">
             </li>
           </ul>
         </div>
-        <table ng-class="{'notimestamp':!settings.showtimestamp,'notimestampseconds':!settings.showtimestampSeconds}">
+        <table>
```
lorenzhs: I would suggest you keep this around and use classes on the `span`s to show/hide instead of `ng-if`s

pmelanson:
Changes made. I haven't run tests on this yet. Reverted to what was happening before with classes, also added in the thing Techman- asked about with zero padding.

lorenzhs:
Thanks, that solved the issue with the watchers. However the right-alignment changes the view a little bit compared to how it was previously. Timestamps are a bit further right in 24h format, creating some wasted space to the left of them. Do you think you could change that?

lorenzhs:
Hmm it misclassifies me (showing 12h format instead of 24h), my browser language settings are (in order of priority, descending): English (UK), English, German, English (US). Got any idea why?

pmelanson:
Removed the right align rule. Had trouble getting chrome to recognize locales, but here's a screenshot of what it looks like now:

no-align

Techman-:
So, what's the status on this?

pmelanson:
I think I've fixed everything, and the original person who asked for this feature is satisfied. I've tested it with US locale (which is 12h) and Canada locale (24h) and it looks like it changes. No real-world testing though.

I think it's ready but that's just me

lorenzhs:
Well then let's just merge it and see if anyone complains :) Thanks @pmelanson and sorry about the delay.

PR #784, opened 2016-06-24
Allow trailing slash at end of gist

pelmers:
Just a one-liner that lets the gist plugin load URL that requires a trailing slash.

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -435,11 +435,11 @@ plugins.factory('userPlugins', function() {
 
     // Embed GitHub gists
     var gistPlugin = new UrlPlugin('Gist', function(url) {
-        var regexp = /^https:\/\/gist\.github.com\/[^.?]+\/?/i;
+        var regexp = /^(https:\/\/gist\.github.com\/[^.?]+)\/?/i;
         var match = url.match(regexp);
         if (match) {
             // get the URL from the match to trim away pseudo file endings and request parameters
-            url = match[0] + '.json';
+            url = match[1] + '.json';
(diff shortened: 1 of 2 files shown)


pelmers:
Doesn't fix the issue, will have to investigate again....

pelmers:
misclick

test/unit/plugins.js
```diff
@@ -134,6 +134,7 @@ describe('filter', function() {
             expectTheseMessagesToContain([
                 'https://gist.github.com/lorenzhs/e8c1a7d56fa170320eb8',
                 'https://gist.github.com/e8c1a7d56fa170320eb8',
+                'https://gist.github.com/bgrins/6595824/',
```
lorenzhs: that test doesn't really test whether the correct embed URL is generated, only whether it matches the regexp

js/plugins.js
```diff
@@ -435,11 +435,11 @@ plugins.factory('userPlugins', function() {
 
     // Embed GitHub gists
     var gistPlugin = new UrlPlugin('Gist', function(url) {
-        var regexp = /^https:\/\/gist\.github.com\/[^.?]+/i;
+        var regexp = /^(https:\/\/gist\.github.com\/[^.?]+)\/?/i;
```
lorenzhs: that doesn't work: 
```js
< 'https://gist.github.com/bgrins/6595824/'.match(/^(https:\/\/gist\.github.com\/[^.?]+)\/?/i)
> ["https://gist.github.com/bgrins/6595824/", "https://gist.github.com/bgrins/6595824/"]
```

Issue #781, opened 2016-05-26
Nick completion in the middle of a line is confused by brackets


kode54:
When attempting to complete a nickname that is surrounded by square brackets, it will work at the start of the buffer, but in the middle of the buffer, it will fail, and instead complete to some nickname that may be vaguely related to the letters inside the brackets, even if they completely match another nickname.

For example, in the #glowing-bear IRC channel right now, all examples lead by whitespace to trigger:

 [ko]<tab> turns into my own name, kode54
 [lo]<tab> turns into lorenzh
 [vo]<tab> turns into oyvindm

PR #786, opened 2016-06-25
fix nick completion within brackets #781

Gibstick:
Changed nick completion to use indexOf instead of search, which makes sense to me because the regex was being escaped anyway.

escapeRegExp is now unused but I left it there in case it's needed in the future. For example, it seems that custom suffix support is possible, and that will need to be escaped.

Manual testing, with cursor at the end of the buffer (note the leading space)
- " [Gib]" no longer tab-completes to Gibstick
- " [A-Za-z]" no longer tab-completes to some nick (I don't know why it picks what it does -- ordering of the nicklist?)
- " [a]" no longer tab-completes to some nick (see above)
- " [fo" will now tab-complete to "[foobar]" if a user "[foobar]" is in the channel, and no longer gives `Error: unterminated character class`

Addresses #781

### DIFF
--- a/js/irc-utils.js
+++ b/js/irc-utils.js
@@ -44,7 +44,7 @@ IrcUtils.service('IrcUtils', [function() {
         var foundNick = null;
 
         nickList.some(function(nick) {
-            if (nick.toLowerCase().search(candidate.toLowerCase()) === 0) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Thanks a lot! @kode54 you might like this ;)

PR #790, opened 2016-07-14
allow users to disable the alt+%d buffer switching

### LINKED ISSUE #572: Option to disable alt keyboard shortcuts
I use alt + [0-9] to switch browser tabs regularly in Chrome. Once I land on the glowing bear tab I'm stuck switching between weechat buffers. It would be nice to have an option to disable any alt keybinding so a user can switch back and forth between tabs without using a mouse or alternative shortcut.

tschuy:
Closes #572. Nice for those of us that habitually use Alt-num to get around Chrome and don't want to get stuck in IRC. :smile:

### DIFF
--- a/index.html
+++ b/index.html
@@ -487,6 +487,16 @@ $ openssl req -nodes -newkey rsa:4096 -keyout relay.pem -x509 -days 365 -out rel
                   </div>
                 </form>
               </li>
+               <li>
+                <form class="form-inline" role="form">
+                  <div class="checkbox">
+                    <label>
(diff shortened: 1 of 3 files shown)


torhve:
Easy workaround is to run gb as a separate window, and treat it as an app, not as a tab. But I do understand that not everyone wants to change their workflow.

lorenzhs:
Guys let's just condition showing the checkbox on non-mobile view (no need to clutter the settings for mobile devices without an alt key) and merge this, I think this is enough of a nuisance to lots of people to add a setting.

lorenzhs:
Superseded by #908 to fix some stuff

PR #791, opened 2016-07-17
Non ascii nick completion

dhcmrlchtdj:
Nick completion doesn't work with CJK characters.

### DIFF
--- a/js/irc-utils.js
+++ b/js/irc-utils.js
@@ -126,8 +126,8 @@ IrcUtils.service('IrcUtils', [function() {
             });
             return coll;
         }, {});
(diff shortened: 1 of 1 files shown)


lorenzhs:
Isn't there a simpler way that doesn't iterate all nicks? This looks like it'd be really slow in busy channels. I mean, what are the requirements for the regex? Which characters do we want to *not match*?

dhcmrlchtdj:
For example, at `#archlinix-cn`
https://ooo.0o0.ooo/2016/07/19/578ef40f7770f.png

There chinese / emoji nickname are not matched.

---

> Isn't there a simpler way that doesn't iterate all nicks?

Maybe `\S`? 😂

Issue #789, opened 2016-07-12
Fetch time format from weechat


lorenzhs:
using https://github.com/glowing-bear/glowing-bear/pull/710 it should be fairly straight-forward to fetch the `weechat.look.buffer_time_format` option. We can't use it directly (it has more syntax, including colours, than JS can handle) but we can definitely use it to distinguish between 24h (%H/%HH) and 12h (%h/%hh) time formats

lorenzhs:
documentation for date format used by angular: https://docs.angularjs.org/api/ng/filter/date

pmelanson:
(relates to #783 #782)

From IRC:

<+lorenzh>	basically we should probably look for %H / %h (translate them to %HH/%hh) and maybe AM/PM markers and add those if present in weechat
<+lorenzh>	if you feel like porting the colour support, knock yourself out, that would be cool - but I think detecting 24h/12h and whether AM/PM markers are used (in 12h) is good enough ™️

I'll get to this at some point, but it won't be immediately.

lorenzhs:
This would also allow us to get rid of the two timestamp options \o/ (see if the weechat time string has seconds shown, and ditch them if not)

PR #792, opened 2016-07-19
789 weechat time format

pmelanson:
WIP PR. I'll go over this sometime soon, but atm it seems to work, at least when I tested it with `weechat.look.buffer_time_format = %H:%M:%S` and `%I:%M:%S` ie 24h and 12h time.

I also touched the fetchConfValue code slightly, adding the ability to pass in a callback function. Let me know if my JS flailings are wrong there.

Addresses #789 

I'll send this to people who brought up the 12h/24h time in the channel.


pmelanson:
This also means we'll be able to remove the timestamp-related settings from the settings page, just haven't gotten around to that

pmelanson:
Currently, my approach is to try and detect whether Weechat's time format is one of (long|short) (12|24) hour time.

An alternate approach is to do a lot of replacing, for example if weechat's time format is called `weechatTimeFormat`,

`weechatTimeFormat.replace("%H", "HH") // 24h hour field`
`weechatTimeFormat.replace("%I", "hh") // 12h hour field`

Instead of our current approach, which just tries to slot the user's weechat configuration into one of the four time formats I kind of arbitrarily defined.

Thoughts on this alternate approach? Our current approach kind of aggressively slots users into one of only four different time formats, but all four are sane formats and we are sure to always slot them into one of these formats.

The alternate approach has the benefit of respecting the user's configuration, but introduces more complexity.

Thoughts?

lorenzhs:
Wow, cool! I think it's better to keep the slot approach, as weechat time format strings can also contain colours and other stuff that would be super hard to parse. I don't quite understand what the `'` do (and I haven't actually tested it) but nice work!

pmelanson:
Oh yeah, the `'` are probably the jankiest thing in here. I think I'm going to have to add a comment explaining them

So you'll notice that some of the strings in there, like `secondsFormat`, have `'`. The `'` are angular's way of delimiting string literals, but having two `'` in a row is a literal `'`. So my first approach was just to surround strings like `timeDelimiter` and `secondsFormat` in `'`, but then when I concatenated two such strings together there would be a stray `'` in the resulting time format. For example, people were reporting `23:23':23`.

TL;DR two `'` in `$rootScope.angularTimeFormat` will show up as a literal `'` in the final timestamp, so we have to be careful about making sure two `'` don't show up together. I'll try to not let it get in the way of the code clarity though, and also add comments to what I have now.

lorenzhs:
Oh right the whole thing is dumped in the `$date` filter. Yeah now it makes sense, thanks.

pmelanson:
Cleaned up the code, there's now no jankiness around `'` and also added comments. Made a new commit that removes the timestamp settings checkboxes.

PR #795, opened 2016-07-28
Update plugin.js

Ninjex:
The Asciinema cast plugin regex was faulty. Casts can have /[a-z]/ characters, not only digits.
Here is a test cast for you to run: https://asciinema.org/a/29qfl1dwsgd25o91nfv3dvvbn

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -400,7 +400,7 @@ plugins.factory('userPlugins', function() {
       * Asciinema plugin
      */
     var asciinemaPlugin = new UrlPlugin('ascii cast', function(url) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Thanks, they probably changed that sometime in the last couple of years since we first wrote that plugin :) However, your regex only checks whether the URL *ends* in a letter or number. Instead it should probably be `([0-9a-z]+)`.

Ninjex:
It's understandable that things change in time! Caught my eye, thanks for supplying me with the correct regex. It has been updated! :)

lorenzhs:
Thanks!

PR #796, opened 2016-07-29
Periodic hotlist sync. Fixes #692

### LINKED ISSUE #692: Sync up unread messages count with weechat every now and then
Right now, we ask weechat for the unread count when connecting, but then only keep track of unread count on our end. If we make a mistake, or if another device connected to the relay reads messages, our unread count will be off.

@torhve on IRC suggested a timer every now and then to sync up unread counts.

how does something like a `setTimeout(aFunction, 60*60*1000)` sound? If so, what general area would that go in?

torhve:
There is a bug in WeeChat when hdata is null, which happens when the
hotlist is empty (meaning all buffers are read). Our websocket callback
machinery expects every command with id to return data, so this code in
current versions of WeeChat will lead to a slow leak. But I think lots
of things in our code will do this already so I'm not entirely sure it's
too problematic to let this patch go by.

We could use infolists instead of hdata, but that is cumbersome to parse
and less performant for WeeChat, and sends more data over the wire.

I propose we make a separate attempt at cleaning up the callbacks. Since
we store time on them we could have a cleanuptask that looks through
them and deletes old callbacks. Maybe @dcormier could have a look?

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -151,21 +151,33 @@ weechat.factory('connection',
                     _requestHotlist().then(function(hotlist) {
                         handlers.handleHotlistInfo(hotlist);
 
-                        if (successCallback) {
-                            successCallback();
-                        }
                     });
+                    // Schedule hotlist syncing every so often so that this
+                    // client will have unread counts (mostly) in sync with
+                    // other clients or terminal usage directly.
+                    setInterval(function() {
+                        if ($rootScope.connected) {
+                            _requestHotlist().then(function(hotlist) {
+                                handlers.handleHotlistInfo(hotlist);
+
+                            });
+                        }
+                    }, 60000); // Sync hotlist every 60 second
+
 
                     // Fetch weechat time format for displaying timestamps
                     fetchConfValue('weechat.look.buffer_time_format',
                                    function() {
+                                       // Will set models.wconfig['weechat.look.buffer_time_format']
                                        _parseWeechatTimeFormat();
-                                   });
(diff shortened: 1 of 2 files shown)


js/connection.js
```diff
@@ -151,21 +151,33 @@ weechat.factory('connection',
...
-                        if (successCallback) {
-                            successCallback();
-                        }
                     });
+                    // Schedule hotlist syncing every so often so that this
+                    // client will have unread counts (mostly) in sync with
+                    // other clients or terminal usage directly.
+                    setInterval(function() {
+                        if ($rootScope.connected) {
+                            _requestHotlist().then(function(hotlist) {
```
lorenzhs: maybe add:
```js
                if (models.version[0] >= 1) {
                    connection.sendHotlistClear();
                }
```

torhve: Why?

lorenzhs: So that the messages that arrived in the current buffer since the last time it had its hotlist cleared are marked as read in weechat

torhve: I'm not sure that is needed or correct always. If it is, we should send hotlistclear for every recieved line that is in an open buffer. I think hotlistclear should always only be sent for user actions.

lorenzhs: Hm we used to just switch buffers in weechat so I tried to mimic the old behaviour but either is fine by me

dcormier:
Really not sure why you wanted me to have a look. ¯\\\_(ツ)_/¯

cormier:
@dcormier I think he wanted me to have a look :)

PR #797, opened 2016-07-30
handlers: check for active buffer

torhve:
Check for active buffer before syncing unreads from WeeChat


lorenzhs:
Wouldn't have happened with clearHotlist :P

PR #799, opened 2016-08-02
Some cleanup based on jshint

torhve:


js/connection.js
```diff
@@ -289,7 +289,8 @@ weechat.factory('connection',
             if (timeout >= 600000) {
                 // If timeout is ten minutes or more, give up
                 $log.info('Failed to reconnect, giving up');
-                handleClose();
+                // This code can't run since handleClose is defined inside connect
+                // handleClose();
```
lorenzhs: we should fix that instead of removing it, I guess

js/filters.js
```diff
@@ -117,7 +117,9 @@ weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
                 }
             }
             // recurse
-            if (node === undefined || node === null) return;
+            if (node === undefined || node === null) {
+                return;
+            }
```
lorenzhs: hmpf I find that style cumbersome if the then-block fits on one line

lorenzhs:
build fails!!!111cos(0)

js/connection.js
torhve: Sure, but that doesn't have to be a part of this PR. And that code wouldn't be able to run anyway so I'm not changing anything really.

js/filters.js
torhve: Yes, it's more cumbersome. But it's easier to maintain, and easier to stick to one style and follow it. Less chance of introducing bugs later.

js/connection.js
lorenzhs: true but at least we now know that there's an issue here :)

js/filters.js
lorenzhs: I don't have a *strong* preference, I just find it unnecessarily verbose and hard to read. But it assumes that everyone knows what they're doing so maybe it's best to not do it.

Would `if (...) { return; }` (in one line) be okay in your opinion? I'm not asking you to change it here but more of a general question

PR #804, opened 2016-08-11
Add underscore to youtube regex

tschuy:
Youtube links can have underscores, like `https://www.youtube.com/watch?v=Xtrsh4U_GCY`. 

Should I add a unit test to check and make sure that videos with _ are recognized? Doesn't look like there's an infrastructure in place in the tests to test that the entire token is the string, but I can add it if you want.


lorenzhs:
Eh I don't think we need a test for that, our tests are far from complete. Thanks for fixing it!

PR #805, opened 2016-08-11
Fix notification count update

torhve:


js/handlers.js
```diff
@@ -359,9 +359,11 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
                 // 1 is message
                 buffer.unread = l.count[1];
                 // 2 is private
-                buffer.notification = l.count[2];
+                // Use += so count[2] or count[3] doesn't overwrite each other
+                buffer.notification += l.count[2];
```
lorenzhs: wait shouldn't that have been an assignment?

PR #807, opened 2016-08-14
Able to resume buffer

JakeStevenson:
This will add the ability to return the the last viewed buffer upon connection.  

When I use glowing-bear on IOS it often closes the connection when switching applications.  I found it annoying to have to choose the buffer every time I launched the application.  This stores the activebuffer in the localstorage and provides a setting that allows the application to automatically set the current buffer back on connection.

### DIFF
--- a/js/bufferResume.js
+++ b/js/bufferResume.js
@@ -1,0 +1,56 @@
+
+/*
+ * This file contains the service used to record the last 
+ * accessed buffer and return to it when reconnecting to 
+ * the relay.
+ */
+(function() {
+'use strict';
+
+var bufferResume = angular.module('bufferResume', []);
+
+bufferResume.service('bufferResume', ['settings', function(settings) {
+    var resumer = {};
+    var key = settings.host + ":" + settings.port;
+
+    // Hold the status that we were able to find the previously accessed buffer
+    //   and reload it.  If we cannot, we'll need to know so we can load the default
+    var hasResumed = false;
+
+    // Store the current buffer as having been accessed.  We can later retrieve it and compare 
+    //   we recieve info from weechat to determine if we should switch to it.
+    resumer.record = function(activeBuffer) {
+        var subSetting = settings.currentlyViewedBuffers;
+        subSetting[key] = activeBuffer.id;
+        settings.currentlyViewedBuffers = subSetting;
+    };
+
+    // See if the requested buffer information matches the last recorded access.  If so, 
+    //   the handler should switch to this buffer.
+    resumer.shouldResume = function(buffer) {
+        var savedBuffer = settings.currentlyViewedBuffers[key];
+        if (!savedBuffer) { return false; }
+
+        if (!hasResumed) {
+            if (savedBuffer === buffer.id) {
+                hasResumed = true;
+                return true; 
+            }
+            return false;
+        }
+    };
+
+    // The handler will ask for this after loading all infos.  If it was unable to find a buffer
+    //   it will need to know so it can pick the default buffer.
+    resumer.wasAbleToResume = function() {
+        return hasResumed;
+    };
+
+    // Clear out the recorded info.  Maybe we'll do this when the user chooses to disconnect?  
+    resumer.clear = function() {
+        record(undefined);
+    };
+
+    return resumer;
+}]);
+})();
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -1,7 +1,7 @@
 (function() {
 'use strict';
 
-var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch'], ['$compileProvider', function($compileProvider) {
+var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'bufferResume', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch'], ['$compileProvider', function($compileProvider) {
     // hacky way to be able to find out if we're in debug mode
     weechat.compileProvider = $compileProvider;
 }]);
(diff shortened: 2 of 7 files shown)


lorenzhs:
Hi Jake,

thanks, that's a very good idea. I'm not really happy about adding a setting, though - I think we can make this work even better without one. We could save the current buffer when switching, and clear it when disconnecting cleanly (i.e. the user wants to disconnect). When connecting, we can check if an old buffer was set, and restore it accordingly. That way, the buffer is only restored if GB was OOM-killed. Maybe we could even add a timestamp when the last action was performed, and only restore if it was killed within the last hour or so.

We should probably also encode the last host:port combo for restore, and only restore if it matches, to prevent messing up if someone uses multiple weechats or the like.

JakeStevenson:
You make some interesting points.  

I see no trouble with clearing the saved buffer when disconnecting cleanly, and only resuming the previous buffer if it attached to the same weechat (in fact, I'm sure it will behave badly with this scenario).  I'll work on that when I get some more time and update the PR.

I'm less sure about the timestamp bit though.  My intention was to always resume the previous buffer.  The fact it forgets when switching apps in IOS is a pain, but I think resuming always is a worthwhile improvement.

Thanks for the feedback, I'll update as soon as I can.

torhve:
What if we just make it default behaviour to always resume last buffer, I could live with that. But maybe we shouldn't update read history, and read marker.

lorenzhs:
Yeah exactly I'm afraid of unintentionally losing highlights and unread status. That's why I suggested to only resume after a "dirty" disconnect. I won't insist on the timestamp bit, I may have overcomplicated that.

js/handlers.js
```diff
@@ -188,11 +188,18 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
                 buffer = new models.Buffer(bufferInfos[i]);
                 models.addBuffer(buffer);
                 // Switch to first buffer on startup
-                if (i === 0) {
+                var shouldResume = bufferResume.shouldResume(buffer);
+                if(shouldResume){
```
lorenzhs: this assignment should probably be inlined into the condition - `if (bufferResume.shouldResume(buffer)) {`

lorenzhs:
I like it - it feels good. Let's just see if anyone complains about the default-switch behaviour? We can always change it to only trigger after dirty disconnects later ;)

js/glowingbear.js
```diff
@@ -159,6 +160,8 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
 
         $scope.bufferlines = ab.lines;
         $scope.nicklist = ab.nicklist;
+        // JAKE 
+        // $scope.settings.activeBuffer = ab.id;
```
torhve: Jake? :)

js/bufferResume.js
```diff
@@ -0,0 +1,56 @@
...
+        settings.currentlyViewedBuffers = subSetting;
+    };
+
+    // See if the requested buffer information matches the last recorded access.  If so, 
+    //   the handler should switch to this buffer.
+    resumer.shouldResume = function(buffer){
+        var savedBuffer = settings.currentlyViewedBuffers[key];
+        if(savedBuffer === 'undefined' || savedBuffer === undefined){return false;}
+
+        if(!hasResumed){
```
lorenzhs: style - we use spaces after `if` and between `)` and `{` in the rest of the project. This line, and the one above/below it.

js/bufferResume.js
```diff
@@ -0,0 +1,56 @@
...
+        var subSetting = settings.currentlyViewedBuffers;
+        subSetting[key] = activeBuffer.id;
+        settings.currentlyViewedBuffers = subSetting;
+    };
+
+    // See if the requested buffer information matches the last recorded access.  If so, 
+    //   the handler should switch to this buffer.
+    resumer.shouldResume = function(buffer){
+        var savedBuffer = settings.currentlyViewedBuffers[key];
+        if(savedBuffer === 'undefined' || savedBuffer === undefined){return false;}
```
lorenzhs: Wouldn't `if (!savedBuffer || savedBuffer === 'undefined')` accomplish the same thing with less verbosity?

js/handlers.js
```diff
@@ -188,11 +188,18 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
...
                     models.setActiveBuffer(buffer.id);
                 }
             }
         }
+        // If there was no buffer to autmatically load,
+        //   go to the first one.
+        if(!bufferResume.wasAbleToResume()){
```
lorenzhs: if style - also, why is the comment on two lines?

js/bufferResume.js
lorenzhs: oh and maybe a line note on why this can be string 'undefined' would be good? Or can it? Maybe the settings module should handle that?

lorenzhs:
I think that's all the pedantry I can come up with for now, thanks a lot for implementing this :)

JakeStevenson:
Rebased and followed lorenzhs's suggestions.

lorenzhs:
Awesome, thanks!

dreamcat4:
So how do we use it? Do I need to clear the cache in my browser?

lorenzhs:
It's live on https://latest.glowing-bear.org/ - you may have to force a refresh (Ctrl+F5 or something, or mash F5 a couple of times)

dreamcat4:
Thank you just tried this. If you are just re-opening the page as a new tab it seems OK. However if you click the 'Power' button to logout --> login (connect button) again. Then it displays the last channel. However theres a couple of issues:

* Channel name is not hilighted (its background color should be different as-if its already been selected).
* Channel window doesnt update when you enter a new message.

Those 2 issues dont seem happen if you've instead opened the url (with auto-login enabled / checked) e.g. from your Bookmarks.

lorenzhs:
See #814 - Jake already fixed it.

PR #810, opened 2016-08-18
Update Electron packager, update PhantomJS to latest phantomjs-prebui…

kode54:
…lt, and update Electron Makefile to fetch the latest version of Electron as of this commit.


lorenzhs:
Thanks a lot, and sorry this took so long. Tested on linux, works ;)

PR #813, opened 2016-09-04
Document how to run everything from one port.

eriktews:
This commit adds some more documentation about encryption to the static landing page of Glowing Bear. It describes how to configure Apache as a server that serves the Glowing Bear static HTML content and also acts a a reverse proxy for the WeeChat WebSocket connection from a single port using HTTPS. This is an advantage for users in networks that do not allow none standard outgoing ports.

### DIFF
--- a/index.html
+++ b/index.html
@@ -183,6 +183,18 @@ $ openssl req -nodes -newkey rsa:4096 -keyout relay.pem -x509 -days 365 -out rel
 /relay sslcertkey
 /relay add ssl.weechat {{ settings.port || 9001 }}
 </pre>
+              <p><strong>Alternative via Apache reverse proxy</strong>: Should you be hosting Glowing Bear and the WeeChat client on the same system, it can be very interesting to use an Apache to serve both, the static HTML content of Glowing Bear and the websocket connection for WeeChat from the same port. The advantage is that users in networks that block some ports are still able to connect as long as port 443 outgoing is permitted.</p>
+<p>First, you need to configure Apache as usual to serve static HTML content via HTTPS and deploy the Glowing Bear files somewhere such as <i>/var/www/html/</i> or a different directory. Make sure that the modules <i>proxy</i> and <i>proxy_wstunnel</i> are loaded by Apache. Then add the following lines to your Apache SSL host configuration:</p>
+<pre>
+ProxyPass "/weechat" "ws://localhost:8001/weechat"
+ProxyPassReverse "/weechat" "ws://localhost:8001/weechat"
+</pre>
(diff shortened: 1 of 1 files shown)


lorenzhs:
Hi @eriktews,

I don't really want to encourage people to self-host Glowing Bear, as it requires manual updating. I think people should only do that if they understand that, so I don't want to advertise it as if it were the recommended way.

However, I do think we should improve our documentation. We currently have a short section on this in the [README's FAQ section](https://github.com/glowing-bear/glowing-bear/blob/master/README.md#faq) - maybe expand that instead?

torhve:
I think this could fit in our wiki, and not in our README, maybe? And a similar entry for nginx proxy.

eriktews:
I assume it's fine in many places, I decided to put it int he landing page since that's where the description how to setup a relay is currently. The FAQ is probably fine too, maybe even the Wiki.

I'm not sufficiently familiar with nginx to write a configuration for that, but I assume it's possible.

lorenzhs:
Maybe a link to the wiki? That would be the most flexible option.

mweinelt:
An nginx equivalent could look like this

```nginx
location /weechat {
	proxy_pass http://localhost:8001/weechat;
	proxy_http_version 1.1;
	proxy_set_header Upgrade $http_upgrade;
	proxy_set_header Connection "upgrade";
}
```

lorenzhs:
Yeah maybe mine is a bit overcomplicated. Also no idea why I'm using https to connect to localhost :D

mweinelt:
The one on the wiki looks great, indeed.

eriktews:
So I still prefer having the documentation in the FAQ since the FAQ is shipped with the source code, the Wiki is not, but having everything in the Wiki is also acceptable.

Another motivation for having nginx or Apache is that when using nginx or Apache in conjunction with let's encrypt, then the certificate can be renewed and reloaded by certbot. This is currently not supported with the weechat client alone.

torhve:
I think a best of both worlds approach would be mention the possibility in the FAQ, especially with letsencrypt as an example for why. And then the FAQ would link to the wiki for specific configuration.

The reason why is we want to avoid the README becoming a huge monstrosity of a technical document.

Another note, you can totally check out the wiki too, as it is markdown documents stored in git, just add .wiki to the end of the repo name when doing git clone.

lorenzhs:
I agree with @torhve. The Let's Encrypt situation is also implicitly mentioned in the Wiki article (no need to rotate certs in more than one place, but without naming Let's Encrypt explicitly)

lorenzhs:
#853 fixes the remaining open things imho.

PR #814, opened 2016-09-07
Fix bug where disconnecting using the 'power' button, resume acted strange

JakeStevenson:
We needed to "reset" the state of the resumer when disconnecting via the power button -- otherwise it would thing it had already done its job.


dreamcat4:
Thanks!

PR #816, opened 2016-09-17
Detect strftime year/month/day specifiers in weechat.look.buffer_time…

csabahenk:
…_format

Accodingly include year/month/day into our time format.


lorenzhs:
Thanks a lot - I modified it a bit to account for ordering (most of Europe uses day-month-year) and added `%e` which is space-padded day-of-month (though I didn't bother to map it to an equivalent, just use zero-padded version, close enough)

PR #828, opened 2016-10-09
highlight buffers with unread messages

xyzz:
This adds a new css class `.active` to buffers which have unread messages. I've also updated the dark theme to make use of it. Here's how it looks like before/after:

highlight-buffers

### DIFF
--- a/css/themes/dark.css
+++ b/css/themes/dark.css
@@ -2138,3 +2138,11 @@ button.close:hover {
 
 
 
+.buffername {
+    color: #888;
+}
+
+.buffername.active {
+    color: white;
+    font-weight: bold;
+}
--- a/css/themes/dark.css
+++ b/css/themes/dark.css
@@ -2142,7 +2142,7 @@ button.close:hover {
     color: #888;
 }
 
-.buffername.active {
+.buffername.unread {
     color: white;
     font-weight: bold;
 }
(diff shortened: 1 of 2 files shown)


Vaesper:
Can back xyzz up on that, just checked and there's no difference in the DOM between "li"s for buffers with and without unread messages (except for the badge nested under them).

Perhaps the name of the class .active should be changed though, it seems to clash with the currently open buffer - change it to .unread instead?

xyzz:
I've changed the class to `.unread` as suggested.

lorenzhs:
What I'm saying is that you don't need to modify index.html at all. Just change the selector for the active buffer to `buffer.active .buffername` and the effect will be exactly the same.

lorenzhs:
Ah, sorry, I misinterpreted. Might not have been fully awake this morning ;)

css/themes/dark.css
```diff
@@ -2138,3 +2138,11 @@ button.close:hover {
 
 
 
+.buffername {
+    color: #888;
```
lorenzhs: contrast is way too low to see anything

lorenzhs:
I don't really like this—we already have the badge to mark buffers with unread messages. Apart from that, contrast for non-new buffers is way too low and unread ones shouldn't be 100% white (we currently use #ddd). @torhve @cormier thoughts?

lorenzhs:
Closing this because #837 is more general

Issue #827, opened 2016-10-09
base16 theme


Vaesper:
So I've made a theme (inspired by [Chris Kempson's base16](http://chriskempson.com/projects/base16/) that can be trivially recolored to fit different preferences. It, however, is not yet in a polished enough state for making part of GB, so I need some help figuring out what to do.

The thing I need help with is how to deal with the .cef-X and .ceb-X where X > 21. The problem is; I have my weechat set up to only use the first 21 colors of my terminal (because I use base16 for my terminal colors as well) and therefore don't have a meaningful opinion on how to deal with the other colors.

Vaesper:
After some bugfixing it works quite well with different palettes. To illustrate: here are some different variant themes that are identical except for the palette defined at the top.

http://pastebin.com/Y677NSsZ: base16 default
http://pastebin.com/4qVgK9WM: the modified base16 default light palette from the original version
http://pastebin.com/i7AX37PY: base16 mocha light theme (made from the mocha theme by flipping the order of the neutral shades)
http://pastebin.com/a8Z6RUm1: modified version of the base16 solarized light palette - the person who converted the solarized palette into base16 did a poor job.

Still need help figuring out how to deal with the 235 unused terminal colours.

Vaesper:
For posterity: decided to go with a simple fallback color for any non-supported extended colors (22-255). If this causes problems for people, they can either consider reconfiguring their weechat to not use these extended colors (recommended) or override the base theme in the custom CSS to define different fallback colors from the palette depending on the original color.

The reason I decided not to define different fallback colors in the base theme is that it's really not well-defined how the different terminal colors "look" compared to the "default" foreground and background colors. Look for example at the different default colors in different terminals; ubuntu terminal has white on purple, white on black and black on white are both common defaults, some people like green on black etc etc. This means that a certain terminal color that renders as darkish shade of cyan can either be a subtly different background color, a subtly different foreground color or a color that's clearly different from both - only dependent on how the user has configured their base colors. What palette color should it be converted to? Nigh impossible to guarantee consistent results across different configurations for different users.

This is why I recommend that users change their weechat config if they wish to use the base16 themes, as it allows them to use different themes on different devices.

lorenzhs:
Ah right, I'm using extended colours. That's probably it. I guess the name prefix "base16-" kinda gives it away that this might not be the best idea ;)

Vaesper:
[...]

PR #829, opened 2016-10-09
Palette-based themes

Vaesper:
Implements [issue 827](https://github.com/glowing-bear/glowing-bear/issues/827)


lorenzhs:
:musical_note: all the nicks are brown... :musical_note: is this intended? For me all nicks have the same colour

Issue #823, opened 2016-10-05
Nicks with background color break in nicklist


Vaesper:
left side is weechat in terminal, right side is glowing-bear:

![](https://i.imgur.com/uhh3Vi7.png)

Inspecting the elements shows that they do not get the appropriate classes applied. The coloring of these same nicks does work in the chat buffers, background color included.

lorenzhs:
Seems to be a problem with `models.nickGetColorClasses`: https://github.com/glowing-bear/glowing-bear/blob/f98dc977371306b12c5492b357ac88db3484b558/js/models.js#L389

Not sure what there is that we're not handling. Any ideas, @torhve @eirikb?

Vaesper:
Just for completeness' sake; the incorrectly classed nicks get "cwf-default" applied to them whereas the correctly classed nicks (ie. the ones without background colour) get a "cef-X" class applied. In the chat area, the nicks with background colour get both "cef-X" and "ceb-Y" classes applied.

lorenzhs:
Hm. Maybe this could be merged with the `parseRichText` function? I don't know enough about formatting in WeeChat, though

oakkitten:
possibly #513 is related

lorenzhs:
I don't think so - there we just use our own because we didn't have a way to get the correct style previously (but now we do, so this could be done). Here, we fail to parse the style correctly.

Vaesper:
I think I see the problem. Colors in weechat can be of the form FG:BG - in the if/else block there is no catch for colors containing the : symbol.

Vaesper:
Unfortunately I don't have a working test setup; I'd be much obliged if someone else could verify that this works (and doesn't break anything).

`function nickGetColorClasses(nickMsg, propName) {
    var colorClasses = [
        'cwf-default'
    ];
    if (propName in nickMsg && nickMsg[propName] && nickMsg[propName].length > 0) {
        var color = nickMsg[propName];

        if (color.match(/^weechat/)) {
            // color option
            var colorName = color.match(/[a-zA-Z0-9_]+$/)[0];
            colorClasses = [
                'cof-' + colorName,
                'cob-' + colorName,
                'coa-' + colorName
            ];
        } else {
            if (color.match(/^[a-zA-Z]+[:$]/)) {
                // WeeChat color name (foreground)
		var fgcolor = color.match(/^[a-zA-Z]+[:$]/)[0];
                colorClasses = [
                    'cwf-' + fgcolor
                ];
[...]

PR #830, opened 2016-10-09
Fixed nicks with background colours breaking in nicklist

Vaesper:
See [issue 823](https://github.com/glowing-bear/glowing-bear/issues/823)

### DIFF
--- a/js/models.js
+++ b/js/models.js
@@ -387,33 +387,45 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
     };
 
     function nickGetColorClasses(nickMsg, propName) {
+        var colorClasses = [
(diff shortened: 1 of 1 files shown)


lorenzhs:
Thanks, this looks good! Can someone confirm that it fixes the issue (I don't have any background color thingies in my config)? You can try it out at https://latest.glowing-bear.org/pull/830/

Vaesper:
Works for me :)

I've tested a mix of different types of colors like "blue:green", "0:1", "blue:0", "0:blue" and even technically invalid - but not filtered out - options like "blue:green:red".

lorenzhs:
Right - soft failure for invalid nonsense is okay I guess ;)

PR #831, opened 2016-10-11
Nick list colors now also work without background color

Vaesper:
Apparently I did not include FG-only colours in my testing of PR 830, this is fixed now.


lorenzhs:
Thanks for fixing this quickly!

PR #835, opened 2016-10-13
Implement alt-h. Fixes #832

### LINKED ISSUE #832: Mark all messages as read?
Hi,

There is a shortcut in Weechat, "Alt+h" on my setup, to mark all buffers as read. I did not find anything equivalent in Glowing Bear and I think it could be really useful.

Typically, I suggest that upon clicking the total number of unread items, a confirm box pops and asks whether we really want to mark all the unread items as read.

What do you think?

torhve:


lorenzhs:
oh god event.key is even worse, it doesn't respect keyboard layout at all

Vaesper:
Tested on FF and Chromium (both on linux). On both it successfully marks the buffers as read but in FF it also opens the Help menu.

In neither case does it sync the state to weechat; in other instances of GB or weechat itself the buffers are not marked read and after a while the unread counts reappear.

lorenzhs:
I don't think it's supposed to sync the state to WeeChat(?)—just local clearing

Vaesper:
But then the read state reverts after a little bit. Can't check right now but I think the buffers stayed marked as read for like a minute before all jumping back up to their old unread message count even if nothing was said in the channel since using alt-h.

torhve:
Oh yeah, I was lazy and only implemented locally. It totally needs to sync back to weechat for it to work reliably.

torhve:
It's just a small matter of adding a line that sends ```/input hotlist_clear`` over relay

lorenzhs:
@torhve using event.key makes our keyboard shortcuts inconsistent. Either we should switch all of them or use `code` here, too.

lorenzhs:
@torhve do you agree with my change? If so, I think we should merge this

torhve:
Go for it!

lorenzhs:
\o/ thanks @torhve @graymalkin!

PR #836, opened 2016-10-13
Change .footer positioning so that it does not jump on iOS

xyzz:
After testing on firefox turns out it's broken, there's now 35px of free space at the bottom. In other browsers I've tested (Chrome, Safari on iOS, and even Microsoft Edge) it works fine however. Not sure what to do here.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -352,8 +352,8 @@ td.time {
 }
 
 .footer {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Hm this doesn't fix the issue with the input bar jumping up by 35px after clicking send on mobile Safari.

PR #837, opened 2016-10-13
Move unread class to buffer <li> instead of ng-if on badge

lorenzhs:
Enables theming, fixes #828 which is now possible with custom CSS


lorenzhs:
@torhve thoughts?

PR #839, opened 2016-10-13
Base16 theme bugfixes and improvements

Vaesper:
- Now featuring a set of different fallback colours for the extended colours instead of just a single one.
- Supports the .unread and .notification classes that will hopefully be merged soon.
- Various bugfixes and tweaks
- Removed mocha light theme.


lorenzhs:
nice! thanks

PR #842, opened 2016-10-14
Fixed horizontal scrollbar issue in firefox mobile

Vaesper:
Please test in other browsers; I am no webdev master and do not have a setup to test for all browsers. Tested under latest FF and Chromium under linux and it doesn't seem to break things there.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -796,6 +796,11 @@ img.emojione {
         padding-bottom: 0;
     }
 
+    #bufferlines > table,
+    #bufferlines > table > tbody {
+        display: block;
+    }
+
     #bufferlines tr.bufferline {
         display: block;
         overflow: hidden;
@@ -822,6 +827,7 @@ img.emojione {
 
     #bufferlines td.message {
         display: inline;
+        overflow-wrap: break-word;
         padding: 0px !important;
     }
 


torhve:
This has been worked on and opened issues for many times, and that css you put in this PR isn't even valid in firefox. Since I last looked at this issue there's been some changes in firefox, seems like https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap overflow-wrap is the one you want in firefox these days.

Vaesper:
Ah, apparently word-wrap is the legacy version of that, you're right. It's not invalid, though, according to the link you sent, just an alternate name for the same property.

Also, the PRs I think you're referring to (the ones that I've seen/seen you respond to) have been about fixing the "word break whenever instead of at sane locations" thing that firefox does, and subsequently cause horizontal scrollbar issues. I completely agree with you that horizontal scrollbars are much worse than words breaking in less-than-perfect places.

However, I am actually fixing the horizontal scrollbars in the mobile layout on FF. The point is that in the mobile layout the .message gets display: inline which makes FF for some inexplicable reason ignore the size of its parent container.

Here, have some codepen.io illustrations:

[Current functionality](https://codepen.io/anon/pen/NRBJRz)
[My proposed fix](https://codepen.io/anon/pen/ALgwLG)
[word-wrap does do its job because things break when it's not there](https://codepen.io/anon/pen/yaqwgw)
[Though overflow-wrap also does it so that's probably better](https://codepen.io/anon/pen/BLPbWA)

Vaesper:
Nevermind, do not merge this. I found the real culprit of mobile layouts breaking, PR coming soon.

PR #843, opened 2016-10-15
Fix mobile layout: horizontal scrollbars (FF) and td alignment

Vaesper:
Okay, once again, please verify that I'm not a complete lunatic and test this on systems I have no access to. Tested: FF 49 and Chromium 53 on linux.

So the HTML fix... Is a complete revert of commit a1f6012fccd6b4d948bc1b2e44d9d80f1b08f4fe. I'm not sure what was going on there, but as far as I can tell there currently is no (and at the time of that commit there was no) `white-space: pre` applied outside of the `td.prefix`.

Allowing the whitespace fixes not only horizontal scrollbars on FF with the mobile layout, but ALSO fixes something I'd noticed in both FF and Chromium; the fact that the prefix is kept on the same line as the message, both UNDER the timestamp when the message starts with a long string. This does not make any sense to me, if there's a long unbreakable string at the start of the message (someone posting a link?) I'd want the timestamp and prefix to be on the same line at least.

Long story short, I have no idea what I'm doing but it seems to work. Pls check.


Vaesper:
Oh and for sake of clarity; the css changes are to "undo" the `vertical-align: top` from the desktop layout. For desktop it's necessary to line up the table cells correctly but in mobile (with its `display: inline`) it looks weird, especially because the time has a smaller font and the prefix a bigger one. To my best estimation probably not a controversial change.

torhve:
When I test this PR I get an extra space between the prefix and the line. Maybe that's what the HTML comments fixes? I don't enjoy that extra space.

lorenzhs:
Hm appears that way, given the wording of #491?

Vaesper:
I think that this whitespace should be attributed to the table-cell padding that is not removed when switching to display: inline. I will tinker with this PR later and include table layout fixes from #842

css/glowingbear.css
```diff
@@ -808,28 +808,37 @@ img.emojione {
     }
 
     #bufferlines tr {
-        padding: 0px 3px;
+        padding: 0 .5em;
         display: block;
+		flex-wrap: wrap;
```
lorenzhs: you're mixing tabs and spaces here - just a heads up :)

Vaesper: argh, stupid editors. thanks for the heads-up

Vaesper:
@torhve you're either going to love me or hate me for what I've done to mobile layout this time :D

Vaesper:
Found some weirdness where the re-ordering of the prefix and timestamp was messed up on some lines on FF mobile. Same bufferline worked fine on FF desktop with the mobile layout, and I have not yet managed to get remote debugging working so not sure what's going on.

lorenzhs:
Hey @Vaesper what's the status on this? I've been using it on mobile (iOS) for a while and really like it

Vaesper:
My weechat box has been offline for a few weeks due to moving and my lack of time to set it up again, but before that I've been using this branch almost exclusively on my phone and never managed to reproduce that weird timestamp/prefix reordering bug (or even see it again).

It needs to be rebased to master (18 commits behind) and checked that it still works, but github reports no conflicts so at least the rebase should just be a fastforward.

Vaesper:
Cannot test it myself right now due to lack of weechat as mentioned previously, but I've gone ahead and rebased the branch. Let me know if things are now (horribly) broken :D

txanatan:
Tested it, and everything works as expected :)

lorenzhs:
I rebased this onto master so I can keep using it on my phone without missing out :)

PR #845, opened 2016-10-18
Implement repeated-prefix, Fix #844

### LINKED ISSUE #844: Feature request: repeated-prefix
Much like repeated-time, but for prefix, so that we can apply conditional styling when the same person is sending multiple messages in a row without interruption by others (for example to hide the prefix).

Vaesper:
Fix #844 

Related to #605 , but not the same as it allows conditional styling in CSS (ie. one can suppress duplicate prefixes in desktop layout but show them in mobile layout).

### DIFF
--- a/js/models.js
+++ b/js/models.js
@@ -365,12 +365,17 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
             });
         }
 
+        var prefixtext = "";
+        for (var pti = 0; pti < prefix.length; ++pti) {
+            prefixtext += prefix[pti].text;
+        }
+
         var rtext = "";
         for (var i = 0; i < content.length; ++i) {
             rtext += content[i].text;
         }
 
-       return {
+        return {
             prefix: prefix,
             content: content,
             date: date,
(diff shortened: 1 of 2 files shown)


Vaesper:
Should conflict with #843 so if either is merged I will have to manually rebase the other.

js/models.js
```diff
@@ -365,12 +365,17 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
...
+            prefixtext += prefix[pti].text;
+        }
+
         var rtext = "";
         for (var i = 0; i < content.length; ++i) {
             rtext += content[i].text;
         }
 
-       return {
+        return {
```
lorenzhs: ehk! Sorry for the pedantry, can you amend this out?

Vaesper: Why? It's aligned properly now!

lorenzhs: Oh it was wrong before. Sorry, didn't catch that. Thanks for fixing in that case :)

lorenzhs:
Thanks!

PR #849, opened 2016-10-31
better swipe handling and nicklist behaviour on mobile

### LINKED ISSUE #838: Feature: toggle nicklist on mobile without going through settings menu.
Ideally implemented to work with swipe and a button, like how channel list works. Sometimes you just want to quickly check who is in a channel.

lorenzhs:
This replaces the "Hide nicklist" setting with an "Always show nicklist setting" (default: false) on mobile. Swiping to the left either closes the bufferlist (if it's open) or opens the nicklist. Swiping to the right either closes the nicklist (if it's open and the above setting is false) or opens the buffer list (if nicklist is closed or the setting is set to true).

This way, it should also work fine in narrow windows on Desktop.

Fixes #838

### DIFF
--- a/index.html
+++ b/index.html
@@ -280,7 +280,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
           </a>
         </div>
       </div>
-      <div id="sidebar" data-state="visible" ng-swipe-left="hideSidebar()" ng-swipe-disable-mouse class="vertical-line">
+      <div id="sidebar" data-state="visible" ng-swipe-left="swipeLeft($event)" class="vertical-line">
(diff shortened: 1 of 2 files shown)


Vaesper:
Is it possible to not use the "Hide nicklist" option for the mobile layout? As in, right now, the mobile layout hides the option but actually changes its value. Instead, it should leave the option as it is (just ignore it) for if someone switches back to desktop layout.

lorenzhs:
Well yes that would be the far superiour solution but honestly I'm too lazy to refactor all of the nicklist showing/hiding logic to account for that...

lorenzhs:
@Vaesper fixed it :)

lorenzhs:
@torhve @cormier @Vaesper feedback?

PR #850, opened 2016-10-31
Spotify: also match other types of embeds (playlists and artists)

### LINKED ISSUE #787: Spotify embeds for playlists
https://developer.spotify.com/technologies/widgets/spotify-play-button/ ← compact layout should be good for our usecase - probably only need to change the regex

lorenzhs:
Fixes #787


lorenzhs:
Depends on #856

PR #852, opened 2016-10-31
Ensure clients are using https

strk:
I understand this may not be wanted by those not having a working https setup, but in case you want to join the "letsencrypt" movement...

I just found it easier to do it like this than battling with apache


lorenzhs:
Thanks, but let's do this with an HTTP redirect (301) instead of JS. This would really complicate local development, for example.

PR #862, opened 2016-11-04
Added cap on max number of unread lines to load

### LINKED ISSUE #859: Slow with buffers which have large numbers of unread messages
Glowing Bear can be quite slugish to load buffers with very large numbers (~4k) of unread messages, and attempts to load all of them in some situations.

Perhaps some option to set the max number of displayed unread messages would be a nice feature?

graymalkin:
This is to improve performance when switching to a buffer with a large
number of unread lines.

Fixes #859

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -145,11 +145,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     $rootScope.$on('activeBufferChanged', function(event, unreadSum) {
         var ab = models.getActiveBuffer();
 
-        // Discard surplus lines. This is done *before* lines are fetched because that saves us the effort of special handling for the
-        // case where a buffer is opened for the first time ;)
-        var minRetainUnread = ab.lines.length - unreadSum + 5;  // do not discard unread lines and keep 5 additional lines for context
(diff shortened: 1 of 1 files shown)


js/glowingbear.js
```diff
@@ -145,9 +145,14 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     $rootScope.$on('activeBufferChanged', function(event, unreadSum) {
         var ab = models.getActiveBuffer();
 
+        // Discard unread lines above 2 screenfuls. We can click through to get more if needs be
+        // This is to keep GB responsive when loading buffers which have seen a lot of traffic. See issue #859
+        var max_unread_lines = $scope.lines_per_screen * 2;
```
lorenzhs: That makes the whole calculation utterly pointless. If you add 10 here and then simplify the formulas a bit you end up with
```js
var linesToRemove = ab.lines.length - (2 * $scope.lines_per_screen + 10);
```
much simpler :)

graymalkin:
@lorenzhs I've simplified as you suggested. Seems to be doing the trick for me :+1:

lorenzhs:
Thanks!

PR #864, opened 2016-11-09
Implement alt-h

### LINKED ISSUE #832: Mark all messages as read?

graymalkin:
Fixes #832 
Fixes #835 

This is a slight change on @torhve's work, and sends the message that is bound to [alt-h in weechat](https://weechat.org/files/doc/stable/weechat_user.en.html).

| Key   | Description                                                            | Command                 |
| ------ | -------------------------------------------------------------- | ------------------------- |
| Alt+h | Clear hotlist (activity notification on other buffers) | `/input hotlist_clear` |

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -420,6 +420,10 @@ weechat.factory('connection',
         }
     };
 
+    var sendHotlistClearAll = function() {
+        sendMessage("/input hotlist_clear");
+    };
+
     var requestNicklist = function(bufferId, callback) {
         // Prevent requesting nicklist for all buffers if bufferId is invalid
         if (!bufferId) {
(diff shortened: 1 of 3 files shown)


lorenzhs:
Thanks! I cherry-picked your commit onto #835. The merge commit is a bit odd, stuff like this happens when you send PRs from `master` and do other stuff as well, which is why I cherry-picked.

graymalkin:
@lorenzhs ah, thanks very much. I'm not the tidiest with my git-foo. :beers:

PR #867, opened 2016-11-10
Fix Alt+[0-9] behaviour to respect order of merged buffers

lorenzhs:
This has been bugging me for a long time. When connecting to more networks, the newest ones will be first in the buffers object, so alt+1 will go there. But it should go to core.weechat. Fix this by adding the index as sort criterion.

This matches the `getBufferQuickKeys()` behaviour.

### DIFF
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -254,23 +254,24 @@ weechat.directive('inputBar', function() {
                     if (($scope.$parent.search.length || $scope.$parent.onlyUnread) && $scope.$parent.filteredBuffers.length) {
                         var filteredBufferNum = $scope.$parent.filteredBuffers[bufferNumber];
                         if (filteredBufferNum !== undefined) {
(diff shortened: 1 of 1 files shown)


js/inputbar.js
```diff
@@ -359,8 +360,7 @@ weechat.directive('inputBar', function() {
 
                 // Alt-h -> Toggle all as read
                 if ($event.altKey && !$event.ctrlKey && code === 72) {
-                    var buffers = models.getBuffers();
-                    _.each(buffers, function(buffer) {
+                    _.each(models.getBuffers(), function(buffer) {
```
lorenzhs: It's kind of ridiculous to fix jshint by getting rid of this binding but honestly I find this solution preferable to renaming `buffers` above. Or maybe pull it up to the beginning of the function?

lorenzhs:
Marked WIP because it doesn't seem to handle filtered buffers correctly

PR #868, opened 2016-11-11
moved repeated-prefix and repeated-time classes to tr

Vaesper:
Also modified the repeated-time expression to apply to a timestamp within 3 minutes of the previous one. If desired, this number can be tweaked.

Does not break the CSS in any of the themes for repeated-time but DOES break the repeated prefix suppression in the wiki. Proposed update for that (with same flaws as current version):

```CSS
.repeated-prefix td.prefix a span {
    position: relative;
    left: -1000px;
}

.repeated-prefix td.prefix a span.hidden-bracket {
    position: absolute;
    left: -1000px;
}

.repeated-prefix td.prefix a span:nth-last-of-type(2):after {
    content:"↪";
    position: relative;
    left: 1000px;
}

.repeated-prefix td.prefix a span.highlight {
    position: initial;
    left: initial;
}

.repeated-prefix td.prefix a span.hidden-bracket:after,
.repeated-prefix td.prefix a span.highlight:after {
    content: initial;
    position: initial;
    left: initial;
}
```


index.html
```diff
@@ -305,13 +305,13 @@ <h4 class="panel-title">
             </tr>
           </tbody>
           <tbody ng-repeat="bufferline in bufferlines">
-            <tr class="bufferline">
+            <tr class="bufferline" ng-class="::{'repeated-prefix': bufferline.prefixtext==bufferlines[$index-1].prefixtext, 'repeated-time': (((60*+bufferline.shortTime.substr(0, bufferline.shortTime.indexOf(':')) + +bufferline.shortTime.substr(bufferline.shortTime.indexOf(':')+1)) - (60*+bufferlines[$index-1].shortTime.substr(0, bufferlines[$index-1].shortTime.indexOf(':')) + +bufferlines[$index-1].shortTime.substr(bufferlines[$index-1].shortTime.indexOf(':')+1)) + 1440) % 1440) < 3, 'highlight': bufferline.highlight}">
```
lorenzhs: This should be an attribute of the bufferline. It's impossible to follow the logic when crammed onto a single line like this

index.html
```diff
@@ -305,13 +305,13 @@ <h4 class="panel-title">
...
               <td class="time">
-                <span class="date" ng-class="::{'repeated-time': bufferline.shortTime==bufferlines[$index-1].shortTime}">
+                <span class="date">
                   <span class="cof-chat_time cob-chat_time coa-chat_time" ng-bind-html="::bufferline.formattedTime"></span>
                 </span>
               </td>
-              <td class="prefix"><span ng-class="::{'repeated-prefix': bufferline.prefixtext==bufferlines[$index-1].prefixtext}"><a ng-click="addMention(bufferline.prefix)"><span class="hidden-bracket">&lt;</span><span ng-repeat="part in ::bufferline.prefix" ng-class="::part.classes" ng-bind="::part.text|prefixlimit:25"></span><span class="hidden-bracket">&gt;</span></a></span></td><!--
+              <td class="prefix"><span><a ng-click="addMention(bufferline.prefix)"><span class="hidden-bracket">&lt;</span><span ng-repeat="part in ::bufferline.prefix" ng-class="::part.classes" ng-bind="::part.text|prefixlimit:25"></span><span class="hidden-bracket">&gt;</span></a></span></td><!--
```
lorenzhs: Does the outermost `span` have any function left? I guess we could just style the `a` it contains instead

Vaesper: I just removed it and it doesn't seem to break anything. Any styling of sub-elements of `.prefix` was already done to `span`-descendants so it applied to the `span`-children of the `a`.

PR #880, opened 2016-12-28
PLS TEST: Fix non-TLS warning

lorenzhs:


lorenzhs:
Tested, works - merging to un-break latest via HTTP

PR #893, opened 2017-02-01
Electron on linux: Use a favicon in place of unsupported app badge

lorenzhs:
Electron doesn't support app badges on Linux.

This patch adds a tray icon when platform === "linux" to display
the highlight counter, as a poor substitute for the app badges.

Snagged from @20after4's fork :)

### DIFF
--- a/electron-main.js
+++ b/electron-main.js
@@ -208,6 +208,22 @@
         mainWindow.loadURL('file://' + __dirname + '/electron-start.html');
         mainWindow.focus();
 
+        /*
+        On linux, update the window icon instead of using the badge feature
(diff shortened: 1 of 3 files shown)


20after4:
I don't think this works, currently.

lorenzhs:
Yeah I think so too

20after4:
I had it working in [electric-gb](https://github.com/20after4/electric-gb/blob/master/main.js) but porting it over to glowing-bear's official electron app didn't work for some reason (still haven't managed to debug it fully)

lorenzhs:
Closed because it doesn't work

PR #894, opened 2017-02-08
Add some image formats

nyuszika7h:
Added support for `.bmp`, `.svg` and `.webp`.


lorenzhs:
That list is terribly incomplete. Thanks for making it a bit less so :)

PR #895, opened 2017-02-08
Added ability to filter out messages (joins/parts/quits/modes)

chetan:
Makes for cleaner display, particularly on mobile


Vaesper:
Why not use something like smart filter in weechat?

chetan:
I'm actually not super familiar with weechat, mostly just using it via glowing bear, but I really only want to enable this filtering on mobile. Other sessions should be unfiltered. Can I do that with smart filters?

Vaesper:
I don't think so, no. There is no way to enable/disable smart filters through GB right now, otherwise that would work.

lorenzhs:
I think we should investigate whether there's a way to show/hide filtered lines dynamically in Glowing Bear. That should make smart filter viable for more people.

lorenzhs:
I'm closing this because it's a bit too nice. Ideally this should be possible in CSS (maybe we could add the classes to the `tbody` object?) but I don't think it should be a setting. Most people will prefer smart filter anyway

PR #897, opened 2017-02-09
[WIP/playing around]Add ability to show filtered lines with Alt+=

lorenzhs:
Still needs lots of work:
- measure performance impact. Non-starter if noticeable.
- fix readmarker. Haven't thought about that at all yet.
- fix toggle jiggle. Should probably keep the topmost line in the same position as before or whatever WeeChat does.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -325,6 +325,13 @@ input[type=text], input[type=password], #sendMessage {
     margin-top: 35px;
     width: 100%;
 }
+#bufferlines > table > tbody.filtered {
+   display: none;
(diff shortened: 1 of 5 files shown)


Vaesper:
How do use this on mobile?

lorenzhs:
Well it does say "WIP/playing around" in the title :) It's not done yet, but I made a PR so anyone can test and play around with it on https://latest.glowing-bear.org/pull/897/

Vaesper:
Sure, just figured I should point out that that's also a thing to consider :)

PR #898, opened 2017-02-09
Add more file formats

nyuszika7h:


lorenzhs:
Can browsers really display all of those? I thought support was pretty limited, especially for audio and video.

js/plugins.js
```diff
@@ -317,10 +317,10 @@ plugins.factory('userPlugins', function() {
     });
 
     /*
-     * audio Preview
+     * Audio Preview
      */
     var audioPlugin = new UrlPlugin('audio', function(url) {
-        if (url.match(/\.(mp3|ogg|wav)\b/i)) {
+        if (url.match(/\.(flac|m3u|m4a|mid|midi|mp3|oga|ogg|opus|pls|wav|wma)\b/i)) {
```
lorenzhs: m3u is just a playlist container, could also be video: https://en.wikipedia.org/wiki/M3U :P

lorenzhs:
Let's just merge and see if anyone complains about their browser not being able to play something that we showed as embeddable

PR #904, opened 2017-02-26
Remove obsolete references to Firefox OS / mozApps

callahad:
Firefox OS has been dead for a year. Might as well remove references from Glowing Bear :(


lorenzhs:
Thanks! I was always surprised how well Glowing Bear worked on the Geeksphone Keon, given the specs of the device. It's sad that Firefox OS didn't work out, but it was a cool idea and Glowing Bear was kind of a nice fit for it. Ah well.

PR #908, opened 2017-03-06
Allow users to disable the alt+%d buffer switching

### LINKED ISSUE #572: Option to disable alt keyboard shortcuts

lorenzhs:
Supersedes #790 

Fixes #572

### DIFF
--- a/index.html
+++ b/index.html
@@ -488,7 +488,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
                 <form class="form-inline" role="form">
                   <div class="checkbox">
                     <label>
-                      <input type="checkbox" ng-model="settings.enableAltSwitcher">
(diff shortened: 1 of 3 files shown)


lilydjwg:
When will this land on https://www.glowing-bear.org/ ? This is the main thing preventing me from using it.

lorenzhs:
@lilydjwg uh we've been a bit negligent with releases, but you can use the version at https://latest.glowing-bear.org/ to get all the latest fixes. I use it all the time.

Issue #910, opened 2017-03-15
Youtube embeds have no title


ttheuer:
I don't know if that was a deliberate choice, but I like to see a video's title before I click play.

the `showinfo=0` at the embed hides the title though. Super easy fix, but I figured I ask before I make a PR: Does this work as intended?

PR #911, opened 2017-03-15
Simplified regex. Added video info.

### LINKED ISSUE #519: YouTube regex is too naïve
Not all YouTube URLs are videos, but the YouTube plugin tries to display an embed for every link to YouTube.

Here's a small sample of URLs that should not be caught but are:
- https://www.youtube.com/youtube
- https://www.youtube.com/user/youtube
- https://www.youtube.com/channel/UCBR8-60-B28hp2BmDPdntcQ
- https://www.youtube.com/yt/about
- https://www.youtube.com/account
- https://www.youtube.com/feed/subscriptions

On legitimate `/watch` video links, if `v` is not the first parameter in the URL then the embed is not displayed correctly. Here's an example URL: https://www.youtube.com/watch?t=32m22s&v=lKkkXInTjNo

ttheuer:
Fixes #519 (as far as I can tell) & #910

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -209,18 +209,18 @@ plugins.factory('userPlugins', function() {
      * See: https://developers.google.com/youtube/player_parameters
      */
     var youtubePlugin = new UrlPlugin('YouTube video', function(url) {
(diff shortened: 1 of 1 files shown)


ttheuer:
Ugh. Forgot to run test -.-
It's 3 AM xD

Issue #909, opened 2017-03-14
Add buffer jump (Alt+J)


lorenzhs:
When pressing Alt+J, GB should:

- Put the search bar in a special mode, indicated visually, maybe with an [Input Group](http://getbootstrap.com/components/#input-groups)
- Display buffer numbers. For this we need to shift the buffer names to the right a bit so that there is enough space
- Filter buffers after first digit has been input
- Immediately jump after two digits have been input (like weechat)

I think this would be a really cool way of implementing buffer jump. I got the inspiration from @ttheuer's [dirty hack](https://github.com/ttheuer/glowing-bear/commit/d38acc7850cd6fe7b1e843da572fc5640f08556e) :)

ttheuer:
*lol* That was quick.

ttheuer:
I added *some* of that functionality: https://github.com/ttheuer/glowing-bear/tree/buffer-jump

Current state:
- [ ] Put the search bar in a special mode, indicated visually.
- [x] Display buffer numbers. For this we need to shift the buffer names to the right a bit so that there is enough space. *Note that this is a clear visual indicator, maybe making the above redundant?*
- [ ] Filter buffers after first digit has been input
- [x] Immediately jump after two digits have been input (like weechat)

ttheuer:
I have no idea what you want to do with searchbar @lorenzhs   
I'll look into filtering the list down.

ttheuer:
I'm probably not going to get to do the other two points, but it's usable already. Should I just create a PR?

lorenzhs:
Uh sure why not but we probably won't merge it yet because it's still work in progress. But it's more visible for anyone to build upon. Whatever you prefer, really

ttheuer:
It might be incomplete but I don't think *I'll* be making any more progress ;)

lorenzhs:
How does WeeChat behave when you press, say, Alt+J 9 and have less than 90 buffers? Does it jump to buffer 9? Could be nice to have I guess

ttheuer:
As far as I can tell, Weechat will just wait until you press a second key and then do nothing.  
Implementing your idea seems frivolous when we have Alt+[0-9] for jumping to single digit buffers.

If users want to enter non-existing buffers they have only themselves to blame if nothing happens!

PR #915, opened 2017-03-15
Alt+J jumps to buffer (like in weechat)

ttheuer:
Partly implements #909 

- [ ] Put the search bar in a special mode, indicated visually.
- [x] Display buffer numbers. For this we need to shift the buffer names to the right a bit so that there is enough space. *Note that this is a clear visual indicator, maybe making the above redundant?*
- [ ] Filter buffers after first digit has been input
- [x] Immediately jump after two digits have been input (like weechat)

The two uncompleted tasks need someone with knowledge of Angular to look at them.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -461,10 +461,10 @@ div.colourbox {
 }
 
 #sidebar .showjumpkeys .buffer .buffer-jump-key {
-    transition: all ease-in-out 0.5s;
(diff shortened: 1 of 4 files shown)


lorenzhs:
I played around with this a bit in https://github.com/glowing-bear/glowing-bear/tree/buffer-jump

lorenzhs:
I'll close this in favour of #922. Thanks so much for doing all of the hard work, @ttheuer!

PR #917, opened 2017-03-15
Replaced 3  no-protocol links with https-links.

ttheuer:
Now they work in the Electron app!


lorenzhs:
Thanks!

lorenzhs:
...if that's what caused them to not work in Cordova I'm going to be quite happy

PR #918, opened 2017-03-17
Add a button to the input bar to complete nicks

txanatan:
Add a button to allow nick completion on mobile, rather than relying on the user using a software keyboard that has a TAB key.

The glyph being used in the button is `glyphicon-chevron-right`, only because that's the only thing in Glyphicons that resembles what most users would perceive as a TAB key.

### DIFF
--- a/directives/input.html
+++ b/directives/input.html
@@ -1,4 +1,4 @@
-<form class="form form-horizontal {{ utils.isMobileUi() ? 'is-mobile-ui' : 'is-desktop-ui' }}" id="inputform" ng-submit="sendMessage()" imgur-drop>
+<form class="form form-horizontal" ng-class="utils.isMobileUi() ? 'is-mobile-ui' : 'is-desktop-ui'" id="inputform" ng-submit="sendMessage()" imgur-drop>
   <div class="input-group">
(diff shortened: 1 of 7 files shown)


lorenzhs:
Thanks! I have a few quibbles, though, not least the space usage which makes the input bar even narrower on mobile. Some issues I noticed:

- button remains in a pressed state (lighter gray background) after pressing it on safari iOS, which looks odd
- it's useless on desktop, we should only show it on mobile
- glyph is easy to confuse for a send button, maybe we can find a better one?

@torhve @cormier thoughts?

txanatan:
Button remaining in a pressed state should be easy to fix, will look into it. Hiding it on mobile should be easy too. The glyph was only chosen because it was the only thing that was available in Bootstrap's Glyphicons, but I'm open to suggestions for changing that to another glyph.

txanatan:
The button isn't remaining in a pressed state - it's just still focused until the touch event ends. This happens with the send button too in my testing. Pushing a commit to hide it on non-mobile now.

directives/input.html
```diff
@@ -1,4 +1,4 @@
-<form class="form form-horizontal" id="inputform" ng-submit="sendMessage()" imgur-drop>
+<form class="form form-horizontal {{ utils.isMobileUi() ? 'is-mobile-ui' : 'is-desktop-ui' }}" id="inputform" ng-submit="sendMessage()" imgur-drop>
```
lorenzhs: `ng-class` is probably what you're looking for

PR #919, opened 2017-03-18
Move Cordova specific fixes into main branch

txanatan:
This makes the Cordova app able to build from the main sources, to prevent having to backport fixes to the cordova branch. 

Settings items that aren't relevant/are disabled on mobile are hidden from the settings menu, and text in the welcome page that refers to browsers has been changed to display the original text on browsers and the Cordova specific text in Cordova (all inline, so it's easy to see what needs changing if the text is updated).

I'm not sure whether the maintainers will want this to be merged or continue with the separated Cordova branch, but this is here for discussion.


txanatan:
The only thing to note is that when building the Cordova app, you must change into the `www/` directory and run `make -f electron.makefile uselocal` before building. Otherwise everything is the same on the Cordova app side.

index.html
```diff
@@ -47,6 +47,7 @@
     <script type="text/javascript" src="js/plugins.js"></script>
     <script type="text/javascript" src="js/imgur.js"></script>
     <script type="text/javascript" src="3rdparty/favico-0.3.10.min.js"></script>
+    <script type="text/javascript" src="cordova.js"></script>
```
lorenzhs: this is something that'll really be better off in a cordova branch (or maybe added during the cordova build process), otherwise it always creates failed requests

index.html
```diff
@@ -63,7 +64,7 @@
         <strong>Connection error</strong> The client was unable to connect to the WeeChat relay
       </div>
       <div class="alert alert-danger" ng-show="sslError">
-        <strong>Secure connection error</strong> A secure connection with the WeeChat relay could not be initiated. This is most likely because your browser does not trust your relay's certificate. Please read the encryption instructions below!
+        <strong>Secure connection error</strong> A secure connection with the WeeChat relay could not be initiated. This is most likely because your <span ng-hide="utils.isCordova()">browser</span><span ng-hide="!utils.isCordova()">device</a> does not trust your relay's certificate. <span ng-hide="utils.isCordova()">Please read the encryption instructions below!</span><span ng-hide="!utils.isCordova()">It's easiest to use a trusted certificate, and remember to include all intermediate CAs. For more info, please read the encryption instructions below!</span>
```
lorenzhs: please use one-time bindings for the cordova checks (prefix with `::`) to prevent creation of watchers. Since this check cannot change at runtime it's safe and will make things faster

index.html
```diff
@@ -248,7 +251,7 @@ <h4 class="panel-title">
         </div>
 
         <div class="title" title="{{activeBuffer().rtitle}}">
-          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky')"></span>
+          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:(utils.isCordova() ? '_system' : '_blank'):{rel:'noopener noreferrer'} | linksForCordova | DOMfilter:'irclinky')"></span>
```
lorenzhs: don't think this will work. I also don't really like evaluating that condition for every line.

index.html
```diff
@@ -314,7 +317,7 @@ <h4 class="panel-title">
...
            --><td class="message"><!--
              --><div ng-repeat="metadata in ::bufferline.metadata" plugin data="::metadata"></div><!--
-             --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' | DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
+             --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | linky:(utils.isCordova() ? '_system' : '_blank'):{rel:'noopener noreferrer'} | linksForCordova | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' | DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
```
lorenzhs: same as above

index.html
```diff
@@ -343,7 +346,8 @@ <h4 class="panel-title">
             <button type="button" class="close" ng-click="closeModal($event)" aria-hidden="true">&times;</button>
             <span class="pull-right version">Glowing Bear version 0.7.0</span>
             <h4 class="modal-title">Settings</h4>
-            <p>Settings will be stored in your browser.</p>
+            <p ng-hide="!utils.isCordova()">Settings will be stored in your browser.</p>
+            <p ng-hide="utils.isCordova()">Settings will be stored on your device.</p>
```
lorenzhs: why suddenly separate the entire `<p>` tag instead of just `<span>`s as above?

index.html
```diff
@@ -499,7 +503,7 @@ <h4 class="modal-title">Settings</h4>
           <div class="modal-header">
             <button type="button" class="close" ng-click="closeModal($event)" aria-hidden="true">&times;</button>
             <h4 class="modal-title">Channel topic</h4>
-            <p ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky')"></p>
+            <p ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:(utils.isCordova() ? '_system' : '_blank'):{rel:'noopener noreferrer'} | linksForCordova | DOMfilter:'irclinky')"></p>
```
lorenzhs: here again

index.html
txanatan: It does work. Probably possible to check once when scope is initialised and then use that variable.

js/filters.js
```diff
@@ -135,6 +135,16 @@ weechat.filter('DOMfilter', ['$filter', '$sce', function($filter, $sce) {
     };
 }]);
 
+weechat.filter('linksForCordova', ['$sce', function($sce) {
+    return function(text) {
+        if (window.cordova !== undefined) {
```
lorenzhs: not sure I like this tbh

index.html
lorenzhs: I think this is unnecessarily complex. It would probably be better to wrap linky appropriately in a new filter

js/glowingbear.js
```diff
@@ -233,7 +240,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         $scope.notifications = notifications.unreadCount('notification');
         $scope.unread = notifications.unreadCount('unread');
 
-        if (settings.useFavico && $rootScope.favico) {
+        if (window.cordova === undefined && settings.useFavico && $rootScope.favico) {
```
lorenzhs: `settings.useFavico` should never be true in cordova so I don't think this is needed

js/glowingbear.js
```diff
@@ -381,7 +397,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     // This also fires when the page is loaded if enabled.
     // Note that this says MathJax but we switched to KaTeX
     settings.addCallback('enableMathjax', function(enabled) {
-        if (enabled && !$rootScope.mathjax_init) {
+        if (window.cordova === undefined && enabled && !$rootScope.mathjax_init) {
```
lorenzhs: Don't think this is needed as `enabled` should never be true in cordova

js/notifications.js
```diff
@@ -200,7 +235,7 @@ weechat.factory('notifications', ['$rootScope', '$log', 'models', 'settings', fu
 
         showNotification(buffer, title, body);
 
-        if (settings.soundnotification) {
+        if (window.cordova === undefined && settings.soundnotification) {
```
lorenzhs: setting should never be able to be true on cordova

index.html
txanatan: Might be a good idea to create a Makefile in the cordova repository then that does the necessary changes before building?

lorenzhs:
Thanks a lot! I really like most of that. I'd suggest using `utils.isCordova()` everywhere instead of inlining the check, though. Also not so sure about the template changes, I'd rather maintain the changes to index.html in a separate branch. I just think that it's easier to maintain and causes less overhead (both in terms of ease of development and machine cycles).

lorenzhs:
Hey @txanatan, what do you think about moving the index.html changes to the cordova branch and keeping the template mostly the same? I'd love to merge this partial approach but I'm a bit unsure about the "full-scale" you propose in this PR.

txanatan:
Sounds good to me, I'll revert all the template changes.

lorenzhs:
I rebased and reverted the template changes

lorenzhs:
This is what the patchset for the new cordova branch could look like: https://github.com/txanatan/glowing-bear/compare/tidyup-cordova...lorenzhs:new-cordova (probably still needs some cleaning)

txanatan:
Looks good to me. I still think that moving 3rdparty stuff in should be done by the Makefile that's already in the repo as part of the Cordova build process rather than committing all of the 3rd party modules into Git though.

lorenzhs:
Ah right, I forgot about that. Thanks for the reminder :)

lorenzhs:
Hey @torhve @cormier could you have a look at this? I think inlining the cordova stuff into the main repo is a great idea that should simplify maintenance significantly.

cormier:
I think it's a great idea as well. I looked at the proposed changes and I find it great such small modifications  would let us build cordova from main branch. I'm not a cordova expert, so please do merge this patch if you consider it ready.

Does it mean that we should be worry about cordova specific stuff when accepting PRs? Maybe we should add something to that regard in CONTRIBUTING.md?

manifest.json
```diff
@@ -9,11 +9,11 @@
   },
   "app": {
     "urls": [
-      "http://glowing-bear.github.io/glowing-bear/"
+      "http://www.glowing-bear.org/"
```
lorenzhs: Eh maybe change this to https?

lorenzhs:
@cormier I think the only thing to consider is that some things are removed in cordova (favico etc), and code referencing that needs to be surrounded by the proper checks (`if (cordova === undefined)` etc).

lorenzhs:
Thanks again @txanatan :)

PR #920, opened 2017-03-18
Changed "close Window" shortcut from Ctrl+W to Ctrl+Q.

ttheuer:
Ctrl+W should be for closing **tabs**, imo. Ctrl+Q would seem to be the standard key for closing a **window**.


lorenzhs:
Sounds fair, thanks :)

PR #926, opened 2017-04-02
[WIP/TEST] Fix nicklist display on iOS

lorenzhs:
Adapted from @Bekcpear

This moves the `nicklist` element *outside* the bufferlines. I added some hacky CSS so that the nicklist is still hidden when the bufferlist is open on mobile but honestly this needs work and probably conflicts with #849


lorenzhs:
Superseded by #989

PR #927, opened 2017-04-05
dev toools: separate shortcut for electron and for web content

20after4:


package.json
```diff
@@ -8,6 +8,8 @@
   "license": "GPLv3",
   "devDependencies": {
     "bower": "^1.3.1",
+    "devtron": "^1.4.0",
+    "electron-debug": "~1.0.1",
```
lorenzhs: what are these and why do we need them?

20after4: I don't remember adding them so I guess it must have been something I was experimenting with that made it's way into the patch.

20after4:
@lorenzhs Ok I got rid of those unrelated package.json changes.

lorenzhs:
Thanks!

PR #928, opened 2017-04-12
Add streamable.com video plugin

JCallicoat:
Support inline previews for streamable.com videos.

Example: https://streamable.com/xh1q1

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -539,8 +539,25 @@ plugins.factory('userPlugins', function() {
         }
     });
 
+    /*
+     * Streamable Embedded Player
+     */
+    var streamablePlugin = new UrlPlugin('Streamable video', function(url) {
+        var regexp = /^https?:\/\/streamable\.com\/s?\/?(.+)/,
+            match = url.match(regexp);
+        if (match) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Hey @JCallicoat, thanks for sending this PR. Is there a particular reason why you're not using `UrlPlugin`? It makes handling URLs a bit easier and handles multiple occurrences, too (so you can have two videos linked in a message).

js/plugins.js
```diff
@@ -539,8 +539,29 @@ plugins.factory('userPlugins', function() {
         }
     });
 
+    /*
+     * Streamable Embedded Player
+     */
+    var streamablePlugin = new Plugin('Streamable video', function(message) {
+        var rPath = /streamable\.com\/s?\/?(.+)/;
```
lorenzhs: this might match unrelated stuff with `streamable.com/blah` anywhere in a URL, it would probably be best to include the protocol explicitly.

JCallicoat:
@lorenzhs I hadn't really looked at UrlPlugin, I will have a go at making the plugin use it!

JCallicoat:
Switched to using `UrlPlugin` and rebased.

lorenzhs:
Thanks!

Issue #903, opened 2017-02-26
Disable automatic emoji in input


SimonSapin:
When I talk about Rust’s `std::io::Write` trait, Glowing Bear automatically replaces `:io:` with `🇮🇴` (U+1F1EE REGIONAL INDICATOR SYMBOL LETTER I, U+1F1F4 REGIONAL INDICATOR SYMBOL LETTER O). If I hit backspace after the moment the replacement happens, the whole sequence is removed. There seems to be no way to input `:io:` literally. This is frustrating.

I disabled the "Enable non-native Emoji support" setting, but that one seems to be about rendering, not input.

I finally manage to disable this "feature" by messing with the source code:

```diff
diff --git a/js/inputbar.js b/js/inputbar.js
index 4777988..80ece0a 100644
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -25,9 +25,9 @@ weechat.directive('inputBar', function() {
                              settings) {
 
             // E.g. Turn :smile: into the unicode equivalent
-            $scope.inputChanged = function() {
-                $scope.command = emojione.shortnameToUnicode($scope.command);
-            };
+//            $scope.inputChanged = function() {
+//                $scope.command = emojione.shortnameToUnicode($scope.command);
+//            };
 
             /*
              * Returns the input element
```

These should probably be a setting for this.

lorenzhs:
I think it's never a good idea to replace matches that aren't surrounded by whitespace, but I don't think emojione has an option to do so. I filed https://github.com/Ranks/emojione/issues/409 a while ago with no response so far.

SimonSapin:
The whitespace thing would be a good start, but regardless I think it’s wrong to make it impossible to input some strings of characters.

Compare with auto-correct on mobile (at least on my phone, Android with Google’s keyboard): when auto-correct changes anything, hitting "backspace" undoes the change and restores what was actually typed. And auto-correction can be disabled completely.

lorenzhs:
Yes, I agree that that would be preferable. I'm not aware of an easy solution, though—EmojiOne isn't very suitable for this usecase :(

SimonSapin:
Proposal:
* Rename the *Enable non-native Emoji support* setting to *Enable non-native Emoji rendering*
* Add another setting called *Enable automatic replacement of EmojiOne short names like `:smile:` in input* or something, that controls whether `emojione.shortnameToUnicode` is used.

[...]

PR #930, opened 2017-04-14
Allow disabling the emoji conversion in the input bar

txanatan:
Simple patch to allow disabling the EmojiOne conversion that happens on inputChanged in the input bar.

### DIFF
--- a/index.html
+++ b/index.html
@@ -474,6 +474,16 @@ npm run build-electron-{windows, darwin, linux}</pre>
                   </div>
                 </form>
               </li>
+              <li>
+                <form class="form-inline" role="form">
+                  <div class="checkbox">
+                    <label>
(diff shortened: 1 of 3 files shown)


lorenzhs:
Hey @txanatan, there's been some discussion about emojification in #903 and to be quite honest I'm currently in favour of just removing it entirely. I don't think many people use it, although I don't have any data to prove it. I was hoping for there to be more of a discussion but participation in decisions has historically been quite low so maybe it would be better to just go ahead with it and see if anybody complains ;) @torhve @cormier say something?

lorenzhs:
this shouldn't be needed any more with #931, @txanatan can you confirm whether everything works as expected on https://latest.glowing-bear.org/?

txanatan:
Yup, everything works as expected. Closing.

lorenzhs:
Brilliant! Thanks for confirming.

PR #932, opened 2017-04-18
Make videoPlugin work for all imgur gifv videos

JCallicoat:
Some gifv links on imgur.com do not have an associated webm video
and only provide an mp4 video. Add two source elements with proper
mimetypes for both types of video. The video player will fall back
to the secondary source if the first cannot be loaded (due to 404)

Example: Trying to load the webm version of this video

  http://i.imgur.com/i7D4GRb.webm

will result in a 302 redirect to the gifv url, which tries to load
an HTML document. This will cause our video player to reject the
document as an invalid video format (text/html); on some videos it
will simply 404. The mp4 version is available and our player will
fall back to that.


JCallicoat:
This can be tested with http://i.imgur.com/i7D4GRb.gifv

lorenzhs:
Thanks! I fixed the https behaviour though

PR #938, opened 2017-05-05
Escape comma in password

### LINKED ISSUE #937: Comma is not supported in password
Since WeeChat 1.6 it has been possible to use a comma `,` in a relay password: https://weechat.org/files/doc/devel/weechat_relay_protocol.en.html#command_init

glowing-bear does not support this and just sends the comma as-is. It should be escaped to attempt to fix it for modern WeeChat versions. Tested this on latest.

lorenzhs:
Fixes #937

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -35,6 +35,8 @@ weechat.factory('connection',
 
             // Helper methods for initialization commands
             var _initializeConnection = function(passwd) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
I'll just merge the easy stuff if nobody reviews...

PR #939, opened 2017-05-05
Fix Ctrl+W when the cursor follows a space.

SimonSapin:
This key binding does "Delete from cursor to previous space". When the text before the cursor was `some example`, it would find the space after "some" and delete "example". When hitting Ctrl+W *again*, it would find the same space again… and delete nothing.

This changes the code to ignore trailing spaces before the cursor for the purpose of finding the previous space, so that something (if at all possible) is always deleted.


lorenzhs:
Thanks @SimonSapin, good catch! It's too bad Chrome doesn't allow us to capture Ctrl-w (it works in Firefox). I expanded the regex to match all kinds of whitespace, not just spaces.

PR #944, opened 2017-05-19
Make buffer search case insensitive

rscullin:
Currently, the buffer search textbox is case sensitive. This is an issue on mobile, where some keyboards automatically capitalize the first letter of a word.

This PR .toLowerCase()s both the buffer name and the search terms, and makes the buffer search case insensitive.


lorenzhs:
I guess it would have been even easier to just use a case-insensitive regex but this looks like it'll do the trick, too. Thanks @rscullin!

Issue #902, opened 2017-02-24
Bad home page user experience when javascript is disabled


itsrachelfish:
When searching for Glowing Bear, the first result is your website [glowing-bear.org](http://www.glowing-bear.org) which I'd expect to contain information about what Glowing Bear is. Unfortunately the site relies on client-side rendering so it's impossible to learn what the site is about without enabling javascript first. Also there's a ton of error messages everywhere.

I understand that Glowing Bear is a webapp and relies on javascript, but don't you think there should at least be some information visible to users that don't have javascript enabled?

![screenshot from 2017-02-24 06 45 19](https://cloud.githubusercontent.com/assets/1670549/23305703/55865b88-fa5d-11e6-8a63-de223d41e164.png)

lorenzhs:
Fair point. Maybe we can just add a `<noscript>` tag that tells non-JS users to check out the github page and excuse the rendering mess? :> It wouldn't be pretty but at least a little bit helpful

Vaesper:
I know for one that uMatrix doesn't trigger `<noscript>` tags.

lorenzhs:
Ugh. But uMatrix users with restrictive defaults should be used to breakage, right? :)

The other option, of course, would be to use something `ng-if="::(!!angular)"` to hide it for js-enabled users. We would just need to prevent flickering.

rscullin:
I've opened a PR ( #945 ) that should fix this issue. Tested and works well for me with uMatrix blocking both 1st and 3rd party scripts.

PR #945, opened 2017-05-19
UX Improvements when JS is disabled or blocked

rscullin:
I've cleaned up the initial GB page when scripts are either blocked or disabled, and now users see an error that they need to fix their settings rather than a jumble of errors and un-templated code. Requested in issue #902.

Preview w/ 3rd party JS blocked:
image


lorenzhs:
Nice! Works well here, too. Thank you @rscullin

PR #946, opened 2017-06-02
Enable title modal on desktop, too, and improve it a bit

### LINKED ISSUE #943: Use modal topic overlay on desktop, too
Otherwise it's impossible to open links that are near the end of the topic, or even comfortably read long topics. The alt text isn't really enough.

lorenzhs:
- make it full-width for clickability of the entire top region
- show the buffer's name in the modal dialog

Fixes #943


lorenzhs:
feedback plz

PR #905, opened 2017-03-01
[DISCUSS] Remove "scroll to readmarker" functionality

lorenzhs:
It's irritating. Fixes #899

PR #950, opened 2017-06-16
[DEBUG] Fix scrolling and simplify logic a little bit

lorenzhs:
I think I broke that in #905

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -211,7 +211,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         });
 
         $timeout(function() {
-            $rootScope.scrollWithBuffer();
+            $rootScope.updateBufferBottom(true);
         });
 
         // Clear search term on buffer change
@@ -582,7 +582,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
             if (bottom) {
                 eob.scrollIntoView();
             }
-            $rootScope.bufferBottom = eob.offsetTop <= bl.scrollTop + bl.clientHeight;
+            $rootScope.bufferBottom = eob.offsetTop <= (bl.scrollTop + bl.clientHeight);
     };
     $rootScope.scrollWithBuffer = function(moreLines) {
         // First, get scrolling status *before* modification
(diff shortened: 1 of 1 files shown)


js/glowingbear.js
```diff
@@ -211,7 +211,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         });
 
         $timeout(function() {
-            $rootScope.scrollWithBuffer();
+            $rootScope.updateBufferBottom(true);
```
lorenzhs: this is the actual fix

js/glowingbear.js
```diff
@@ -582,7 +582,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
             if (bottom) {
                 eob.scrollIntoView();
             }
-            $rootScope.bufferBottom = eob.offsetTop <= bl.scrollTop + bl.clientHeight;
+            $rootScope.bufferBottom = eob.offsetTop <= (bl.scrollTop + bl.clientHeight);
```
lorenzhs: just for clarity, operator precedence means this doesn't change anything

js/glowingbear.js
```diff
@@ -599,12 +599,8 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
                 if (moreLines) {
                     // We fetched more lines, keep the scroll position constant
                     bl.scrollTop = bl.scrollHeight - bl.clientHeight - sVal;
-                } else {
-                    // New message, scroll with buffer (i.e. to bottom)
-                    var eob = document.getElementById("end-of-buffer");
-                    eob.scrollIntoView();
```
lorenzhs: this basically does nothing that passing `true` to the subsequent `updateBufferBottom` call doesn't, so do that instead.

lorenzhs:
Still not good. Just had this:
```
======= activeBufferChanged to #emacs
glowingbear.js:221 activeBufferChanged rAF calling uBB(true)
glowingbear.js:594 updateBufferBottom( true ) -> false
glowingbear.js:167 activeBufferChanged nicklist handler call sWB
glowingbear.js:602 scrollWithBuffer( undefined )
```
The second of those lines should have a `true` at the end. Drawing hasn't finished apparently and it's not scrolling.

js/glowingbear.js
```diff
@@ -197,7 +197,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
                         $rootScope.updateBufferBottom(true);
                         $rootScope.scrollWithBuffer();
                         bl.onscroll = _.debounce(function() {
-                            $rootScope.updateBufferBottom();
+                            $rootScope.updateBufferBottom($rootScope.bufferBottom);
```
lorenzhs: OK this is nonsense, it makes it (nearly) impossible to scroll away from the bottom. FFS.

lorenzhs:
If anyone wants to continue this, I would welcome it very much, but the event chain here requires a deeper understanding of what happens in the browser than I possess.

PR #952, opened 2017-06-26
Update travis' node version

lorenzhs:


lorenzhs:
it would probably speed up the tests a lot if we removed dependencies like emojione (why is it even in package.json?)

PR #959, opened 2017-07-30
Upgrade karma-phantomjs-launcher to 1.0.0

TPXP:
`phantomjs` was renamed to `phatomjs-prebuilt` but this change was integrated in the 1.0.0 release of `karma-phantomjs-launcher`. The previous versions import the `phantomjs` module, which results in a failure since this module is not anymore installed. Consequently, the `run_tests.sh` script hands infinitely because the launcher fails.

This commit makes the run_tests.sh script work again. :-)


lorenzhs:
Thanks!

PR #960, opened 2017-07-30
Add a link to a PKGBUILD for Archlinux users

TPXP:
It's always cool to have a `PKGBUILD` for every app you use on Arch Linux. I created one for glowing-bear, you might want to add a link to it to the project's README to make the setup easier. 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ You can run Glowing Bear in many ways:
  * Chrome app ("Tools", then "Create application shortcuts")
  * Android Chrome app, a full-screen experience ("Add to homescreen").
  * [Android app](https://play.google.com/store/apps/details?id=com.glowing_bear) that you can install from the Google Play Store
- * Electron app, for Windows, Linux and MacOSX. ```npm install; npm install electron-packager; npm run build-electron-{windows, darwin, linux}```
+ * Electron app, for Windows, Linux and MacOSX. ```npm install; npm install electron-packager; npm run build-electron-{windows, darwin, linux}```. If you're using Archlinux, a [PKGBUILD](https://aur.archlinux.org/packages/glowing-bear-git/) is available.
 
 <a href="https://play.google.com/store/apps/details?id=com.glowing_bear"><img alt="Android app on Google Play" src="/assets/img/badge_playstore.png" /></a>
 


lorenzhs:
How do updates work with that? Do you have to push a new version of the PKGBUILD if there is a change in the repository or will it track master and update whenever it changes? (Sorry, I don't use Arch)

0x52a1:
It gets the latest master from git when the package is built, so it's up to the user to update the package if they want new features.

lorenzhs:
Hm I'm not sure I like that, it's very easy to get stuck with an outdated version that way. It's not only about features, it's also about bugfixes and (very rarely, fortunately) security updates. I don't like linking to something that isn't updated, especially if it's not clearly visible that that is the case.

On the other hand, automating the whole electron thing is kinda nice.

TPXP:
The problem is the same if the user installs it following the instructions in the Readme. You might want to display a warning about an updated version on the homepage to warn about outdated versions (eventually use a red background for security updates)

lorenzhs:
But if you clone a repository you don't expect to receive updates
automatically. With a package in your package manager, there is an
expectation of receiving updates. That's a big difference for me.

TPXP:
Hello !

It's been a while, I apologize for not answering this important remark. I believe most AUR users know that AUR packages are not updated, and that they have to rebuild them for every update. AUR package managers like [trizen](https://aur.archlinux.org/packages/trizen/) make it a bit easier by querying the AUR about each AUR package's version but that's not a very good solution for git packages because updates are too frequent.

Other projects have similar problems because you can't expect packages to be updated right after they're out. Gitlab Community Edition addresses this by displaying a square in the upper right corner with the update status. I can show 3 values:
* update asap (red - security fixes)
* update soon (orange - new version)
* up to date (green)

We could implement something similar by making the code "phone home". Consider the following implementation:
* In the repository, a `security_revision` contains a number, increased every time a security fix is applied
* Since this repository is deployed with Github pages, the latest version of this file will be available at https://www.glowing-bear.org/security_revision
* Every client makes a request to get https://www.glowing-bear.org/security_revision on startup and compares it with its own. If lower, it shows an alert and asks to update

I believe this addresses the security concerns, let me know what you think of it.

lorenzhs:
I don't want to add code that checks for updates because it would mean having centralised infrastructure, and we pride ourselves on not requiring any of that.

Maybe I just disagree with Arch's philosophy. I don't know. @cormier @torhve what do you think?

PR #965, opened 2017-08-30
Recognize /raw/ pastebin URLs

tribut:
This fixes the "show pastebin" button when people link directly to the raw version.


lorenzhs:
Thanks! A non-capturing group might have been even nicer ;)

PR #967, opened 2017-09-25
Tweak to the CSS for the nicklist

graymalkin:
This removes the funny empty space at the bottom of the nick list when it is scrollable.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -273,7 +273,7 @@ input[type=text], input[type=password], #sendMessage {
     top: 0;
     margin-top: 39px;
     padding-left: 5px;
-    padding-bottom: 35px;
+    padding-bottom: 39px;
     z-index: 2;
 }
 #nicklist ul {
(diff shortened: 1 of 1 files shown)


lorenzhs:
It kinda removes a bit too much space, don't you think? The last nick ends flush with the bottom of the page. There should be at least a couple of pixels of space there imho

graymalkin:
You're quite right, I've modified the padding at the bottom to match, so it now looks a lot better.

lorenzhs:
I've played around with the values a bit and I think `padding-bottom: 44px` looks better and also matches the effective spacing at the top (I'm unsure how exactly this adds up, but when I hover over the first nick and observe the amount of space to the top and do the same with the last one, 44px works best for me). What do you think?

graymalkin:
Yeah looks good to me! Much nicer than the text touching the bottom of the frame.

lorenzhs:
Thanks, I fixed it up and merged it :)

PR #973, opened 2017-11-11
Fix/Update Electron Windows Build

rinyre:
I added a favicon.ico based on the glowing-bear.icns file, and changed the Windows build to use the x64 platform. Windows x64 is so common now that there's little reason to depend the ia32 build. 

I also removed ASAR because it was causing a crash on exit. While this does remove having a nice single file for the app it does remove the error message on exit.


lorenzhs:
@torhve this one is for you to look at

Fmstrat:
Hey guys, I'm doing the change to make a `build` and `dist` folder so the electron build doesn't upset the git structure (by modifying index.html and such like it does now). In doing so, I've incorporated this PR to get the windows build working.

Fmstrat:
This is the commit that integrates this PR. It is the first commit in implementing the GitHub Action for building pre-compiled Electron releases into each Release. It will come in a PR with the action setup. https://github.com/Fmstrat/glowing-bear/commit/0da87dd8f6bdb84574452035d624e11bd19abae3

PR #977, opened 2018-01-11
extend emojify filter to convert shortnames as well

sn0opy:
This change will convert shortnames like `:see_no_evil:` (🙈) to emojis / images as well. Emojione will convert only known shortnames to its equivalent. Thus custom Slack emojis will be displayed as text.

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -180,8 +180,10 @@ weechat.filter('emojify', function() {
             // Emoji live in the D800-DFFF surrogate plane; only bother passing
             // this range to CPU-expensive unicodeToImage();
             var emojiRegex = /[\uD800-\uDBFF][\uDC00-\uDFFF]/g;
-            if (emojiRegex.test(text)) {
-                return emojione.unicodeToImage(text);
+            // regex to look for typical shortname syntax in tools like Slack
+            var shortnameRegex = /\:[\+-_a-zA-Z0-9]+\:/g;
+            if (emojiRegex.test(text) || shortnameRegex.test(text)) {
+                return emojione.toImage(text);
             } else {
                 return(text);
             }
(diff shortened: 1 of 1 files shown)


oakkitten:
```var shortnameRegex = /\:[\+-_a-zA-Z0-9]+\:/g;```

`+-_` includes stuff like `[` and `/` and the rest of characters in this regex

sn0opy:
@oakkitten nice find. fixed.

lorenzhs:
I'm afraid this would only make the emojification issues worse, our input bar replacement already breaks a few things (like mac addresses that happen to contain :de: etc) and this would apply that to *everyone's messages*, not just stuff that you're typing where you see the problem in real time.

It's a nice feature when it works but I think we should consider the potential issues carefully

sn0opy:
@lorenzhs without looking further into it, but one could extend the regex to force a white space before and after the shortname syntax. For messages which contain just an emoji one could easily add an exception to it.

lorenzhs:
Unfortunately it's not that easy because emojione will still replace the `:io:` in `lala :af: std::io::blub` and yield lala :afghanistan: std:🇮🇴:blub`... Check out https://github.com/glowing-bear/glowing-bear/blob/f47043fe177db172a7a98386e0877d031c742597/js/inputbar.js#L33 for the gymnastics required to fix this. We should probably factor that code out and use it for both.

oakkitten:
why not look for the patterns and replace only those? sorry don't know js so python

    input = ":one: :two: :three:four: :five:six:"
    r = re.compile(r"(?:^|(?<= )):[-_+a-zA-Z0-9]+:(?= |$)")
    r.sub(lambda m: m.group().strip(":"), input)

yields 

    'one two :three:four: :five:six:'

lorenzhs:
It's a bit more complicated for the input bar because the replacement shouldn't change the logical cursor position, but for the text lines, it might be enough. We should measure performance though to avoid a potential slowdown.

PR #981, opened 2018-02-22
Update README.md

vectr0n:
Add note about being able to serve glowing-bear from a web server without the need to proxy, also added an nginx example.


lorenzhs:
Hi @vectr0n,

I'm not sure I quite understand you. The readme states rather clearly that "[a]ll you have to do is clone the repository, fire up a webserver to host the files, and start fiddling around." There's no proxying of any kind going on when you use our hosted version at https://www.glowing-bear.org or https://latest.glowing-bear.org (I would recommend the latter, btw, it's much better – we need to make a release), and you always get the latest version without having to update your own copy.

When we talk of proxying, what we mean is proxying the *relay connection* behind nginx (or apache, etc). That means not having to expose WeeChat directly to the outside world, but rather letting nginx take care of terminating the connection and handling things like TLS or rate limiting. In such a setup, nginx would then pass on an unencrypted connection to WeeChat, which means that you wouldn't have to care about things like certificate renewal in WeeChat. It's easier to maintain this way.

So I'm really not sure what you mean. Your nginx configuration is a completely standard way of hosting static files, so I'm not sure what it would add.

lorenzhs:
I hope that I could clarify things a bit :)

PR #986, opened 2018-03-17
Fix for jumping input bar in Safari

forberg:
Fixes this issue: https://github.com/glowing-bear/glowing-bear/pull/836#issuecomment-253514624

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -384,7 +384,7 @@ td.time {
 }
 
 .footer {
-    position: fixed;
+    position: sticky;
     bottom: 0;
     height: 35px;
     width: 100%;


lorenzhs:
Hm testing on mobile safari this messes up page height adjustment but fixes some issues I’ve been having with the cursor being misaligned in the input bar on iOS 11. The window still jumps to the top when I send a message (more precisely, when the message I sent is received again).

lorenzhs:
Now I’ve still got the jumpiness and the input bar blocks more space than it needs (it extends too far up). Got any ideas how to fix that?

Thanks btw for taking a shot at our issues with Safari, it’s much appreciated! :)

lorenzhs:
Hey @forberg, I just tried this in Chrome 65 on the Desktop and it messes up the input bar placement (it's floating a bit too high but a little different in every buffer, it seems). Are you sure that `position:sticky` is the right thing here?

PR #987, opened 2018-03-18
Fix tests and update npm dependencies

lorenzhs:
- Updates node version on Travis from 0.10 (!!) to 8
- Updates a bunch of npm dependencies for the tests to more recent versions
- Splits bower dependencies into two sets:
  - those needed for tests (`bower install -p`)
  - the rest, needed for full offline resources, i.e. electron build (`bower install`)
- Also fixes the build by removing an unnecessary semicolon :smile: 

Supersedes #952


lorenzhs:
I'm just going to go ahead and merge this because it only affects development infrastructure, not what anyone is running in their browser when they open https://latest.glowing-bear.org/

PR #988, opened 2018-03-18
Update electron and electron-packager to latest version

lorenzhs:
Tested on Linux, @torhve can you check if everything works on windows?


lorenzhs:
I'll just merge this, it's low-impact if something breaks and probably fixes a bunch of security issues in electron

PR #989, opened 2018-03-18
Fix nicklist on mobile: now works on iOS & behaves correctly buffers without a nicklist

### LINKED ISSUE #924: nick list cannot display on iPad
I found this bug and try a way to fix it.

The `#nicklist-div` is in the `#bufferlines-div` by default. When nick list displaying under my iPad Safari, its content is covered by `margin-right` of `#bufferlines-div`.  I also try to modify the z-index property but have no effect.

If move the `#nicklist-div` to outside of `#bufferlines-div`, everything looks fine. Check the index.html in this commit: https://github.com/Bekcpear/glowing-bear/commit/74992279a0db4781cb226a9068b2299e5b9527f8

*I have no idea whether this modification can affect other features, and I also have only iPad(with version 10.2.1) so I don't know if it works on iPhone or other iOS device.*

lorenzhs:
Fixes #924 \o/

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -356,7 +356,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         } else if ($scope.swipeStatus === 0) {
             // show nicklist
             $scope.swipeStatus = -1;
-            $scope.updateShowNicklist();
+            if (!$scope.updateShowNicklist()) {
+                $scope.swipeStatus = 0;
+            }
         } else if ($scope.swipeStatus === -1) {
             /* do nothing */
         } else {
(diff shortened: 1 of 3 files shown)


lorenzhs:
The iOS fix was originally adapted from @Bekcpear, thanks!

js/glowingbear.js
```diff
@@ -761,26 +763,30 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
-    // Utility function that template can use to check if nicklist should
-    // be displayed for current buffer or not
-    // is called on buffer switch and certain swipe actions
+    // Utility function that template can use to check if nicklist should be
+    // displayed for current buffer or not is called on buffer switch and
+    // certain swipe actions.  Sets $scope.showNicklist accordingly and returns
+    // whether the buffer even has a nicklist to show.
     $scope.updateShowNicklist = function() {
-        $scope.showNicklist = (function() {
+        // returns array of booleans: [show nicklist, buffer has nicklist]
```
lorenzhs: this should probably update $scope.showNicklist and return the value directly, without wrapping in another function and doing the stupid wrapping dance.

lorenzhs:
So uh I'll just merge this and see if anyone complains

PR #990, opened 2018-03-18
Don't put hidden angle brackets around actions and other non-messages

### LINKED ISSUE #616: Don't put invisible <> for copy&paste around non-chat prefixes such as join/quit messages

lorenzhs:
Fixes #616

### DIFF
--- a/js/models.js
+++ b/js/models.js
@@ -359,6 +359,11 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
         var highlight = message.highlight;
         var content = parseRichText(message.message);
 
+        // only put invisible angle brackets around nicks in normal messages
+        // (for copying/pasting)
+        var showHiddenBrackets = (tags_array.indexOf('irc_privmsg') >= 0 &&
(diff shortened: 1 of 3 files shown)


lorenzhs:
I don't think this should be too expensive (computationally), but I haven't benchmarked it.

js/models.js
```diff
@@ -359,6 +359,11 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
         var highlight = message.highlight;
         var content = parseRichText(message.message);
 
+        // only put invisible angle brackets around nicks in normal messages
+        // (for copying/pasting)
+        var showHiddenBrackets = (tags_array.indexOf('irc_privmsg') >= 0 &&
+                                  tags_array.indexOf('irc_action') === -1);
```
lorenzhs: are these the right classes?

PR #992, opened 2018-03-22
Fix electron build

zakius:


electron.makefile
```diff
@@ -13,12 +13,12 @@ copylocal:
 
 # modify index.html to use local files
 uselocal: copylocal
-	sed -i.bak 's,https://cdnjs.cloudflare.com/ajax/libs/[^\"]*/,3rdparty/,g' index.html
+	sed -i.bak 's, https://cdnjs.cloudflare.com/ajax/libs/[^\"]*/,3rdparty/,g' index.html
```
lorenzhs: This isn't correct? There is no space preceding the URL in `index.html` so making this change prevents the replacement from being applied.

lorenzhs: You need to run `npm run make-local`, not `make -f electron.makefile uselocal` (which doesn't download the required files), or run `bower install --dev` manually to download them.

lorenzhs:
The rest is a duplicate of #973 which is waiting for review by @torhve

electron.makefile
zakius: I downloaded files via bower and then executed npm run build-electron-windows according to readme

Building like that caused sed to throw 
>sed: -e expression #1, char 63: unknown option to `s'

adding space caused build to go forward and get stuck on setting icon, quick research told me it is upstream issue with rcedit common for electron projects, working around that caused build to success and output runs just fine except crash on exit that indeed is addressed in #973

running `npm run make-local` throws the same sed error

PR #800, opened 2016-08-03
Ability to pin buffers. Fixes #778.

torhve:
This introduces ability to pin buffers using localvar storage in
WeeChat. At a later stage some UI should be introduced for this, like
discussed in #604 .

PR #998, opened 2018-04-02
Give an id to buffers se we can play with them with custom CSS.

JulienPalard:
I admit, my usage is:

    #buffer-irc\.freenode\.\#some_useless_chan {display: none;}

But I imagine some may have more sound use cases.

### DIFF
--- a/index.html
+++ b/index.html
@@ -300,6 +300,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
                                         'channel_plus': buffer.prefix === '+',
                                         'channel_ampersand': buffer.prefix === '&',
                                         'private': buffer.type === 'private'}"
+              ng-attr-id="{{ 'buffer-' + buffer.fullName }}"
               ng-repeat="(key, buffer) in (filteredBuffers = (getBuffers() | toArray:'withidx' | filter:bufferlistfilter | filter:hasUnread | orderBy:predicate | getBufferQuickKeys:this))">
             <a ng-click="setActiveBuffer(buffer.id)" title="{{ buffer.fullName }}" href="#">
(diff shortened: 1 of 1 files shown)


lorenzhs:
Hm, what if we added a setting to hide a buffer *unless* it's searched for? Similar to what we do with pinning (see "Pinning buffers" under usage instructions in https://latest.glowing-bear.org/). That way the buffer isn't *completely gone* but just hidden unless you're searching for it, and it would complement the existing pinning feature nicely.

JulienPalard:
@lorenzhs I'd love the "hide buffer unless it's searched for" feature! But I don't feel like implementing it any time soon (My front-end-fu is no good).

lorenzhs:
I think it should be rather easy to do, given how easy pinning was (#800). I might have a stab later.

lorenzhs:
Wait, weechat already supports hidden buffers, right? Maybe we should just reuse that property

lorenzhs:
Eh it looks like this should already work! Just mark the buffer as hidden in weechat https://github.com/glowing-bear/glowing-bear/blob/005f9f6f5ce3be837fc82f038b0ae62712a26363/js/glowingbear.js#L730

JulienPalard:
Wah nice! I'm testing it right now, I think its worth mentionning in the usage that glowgingbear handle this (which I can do after testing it properly).

JulienPalard:
It works almost perfectly, only bug I see is jumping using `Alt-a` also jumps to hidden buffers.

lorenzhs:
@JulienPalard can you try https://latest.glowing-bear.org/pull/999/ in ~15 minutes? (need to wait for sync)

lorenzhs:
I'll close this because it seems that it's not actually needed.

PR #1001, opened 2018-04-13
Make image uploadable by paste from clipboard

jungeonkim:


lorenzhs:
Good idea, would be nice to have!

lorenzhs:
Works like a charm, I like it. Thanks!

Issue #1002, opened 2018-04-21
Jump to hidden buffer if it contains a mention.


JulienPalard:
This follows https://github.com/glowing-bear/glowing-bear/pull/999:

When I'm getting a mention on a hidden channel, hitting Alt-a does not jump on it. So I have to guess on which chan I got the notification to search it by name to find it, there is no other mean of finding the mention.

Steps to reproduce:

- Create a test channel and join it with latest.glowing-bear.org.
- Hide it (`/buffer hide`)
- Switch to another buffer
- Join your hidden channel from webchat.freenode.net
- Mention the nick you use on glowing-bear from freenode
- Switch back to glowing bear
- See the red "1", but there is no way to jump on it other than searching it by name.

PR #1003, opened 2018-04-21
jump to buffers with notifiction with Alt-A, even if hidden.

JulienPalard:
Playing around https://github.com/glowing-bear/glowing-bear/issues/1002


lorenzhs:
lgtm, thanks!

PR #1017, opened 2018-07-07
Fix bufferlist item height on condensed layout

### LINKED ISSUE #1012: style/theme issue: channels with messages
In the channel list on the left, when new messages happen in a channel an indicator with the number of messages appears. this indicator makes the box slightly taller than the rest of the entries.

I think they should all be the same height.

lorenzhs:
Fixes #1012, based on @rain-1's suggestion

This slightly increases the height for the other buffers, but that's required to prevent jumping.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -851,6 +851,11 @@ img.emojione {
         font-size: 14px;
     }
 
+    .nav-pills li.buffer {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Honestly I'll just merge because it's a trivial change

PR #1020, opened 2018-08-19
Keep track of unread counts for servers' buffers for 'show only unread' view

### LINKED ISSUE #1019: also not show *server* buffers if no unread
Would be great if setting "only show buffers with unread messages" would also hide server buffers (when no unread in any of server's related buffers, perhaps)

lorenzhs:
Maintain unread info for all buffers belonging to a server.  This info
is used to hide server buffers for which *no* associated buffer has
activity when 'Only show buffers with unread messages' is enabled.

Closes #1019


quite:
Hm, one trick this might create, is that it would be tricky to use a server (to join a chan, say)/which have no buffers, or where they are all hidden due to inactivity. Perhaps turning off this display preference is a workaround.

lorenzhs:
Searching for a buffer should still work, although I haven’t tested it

quite:
Yes, searching works!

Issue #514, opened 2014-11-30
Handle _buffer_moved and _buffer_merged


oakkitten:
newly opened buffer is shown at incorrect position, i suppose it fails to get merged

_Edit (lorenzhs):_ the `_buffer_moved` and `_buffer_merged` signals aren't handled at all at the moment. This causes the behaviour described above. We should add handlers for those.

PR #1023, opened 2018-10-05
Handle buffer_moved

dkess:
This fixes part of  #514, by handling buffer_moved. It goes through the list of buffers and adjusts their numbers to shift them left or right based on how the buffer was moved.


lorenzhs:
Thanks! Looks good to me.

PR #1024, opened 2018-10-06
Add a Dockerfile and .dockerignore

bl1nk:
This adds a simple `Dockerfile` to run glowing-bear on Docker.  
I was unsure if more files can be added to the `.dockerignore`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.*
+*.sh
+*.md
+test
+electron-*
+COPYING
+package.json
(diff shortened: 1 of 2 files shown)


lorenzhs:
Uh, I'm not quite sure what the purpose of docker is for a project that consists of a bunch of static files. For development, you're much better of using `python3 -m http.server`, and for deployment, you're probably already running a web server somewhere, and forwarding requests to the docker container would be more work than just serving the files in the first place. Am I missing something?

bl1nk:
@lorenzhs You're not missing anything. Having an official container on Dockerhub would be nice to have. You can automate Dockerhub to publish a new container on every release.

All of this would just be convenience. I'm fine if this is not desired.

PR #1025, opened 2018-10-06
[WIP] Respect safe area constraints on webkit

bl1nk:
This adds limited support to respect constraints on newer iPhones, especially when used as PWA.

More information about this can be found on [the official webkit blog](https://webkit.org/blog/7929/designing-websites-for-iphone-x/).

There's one more issue I am facing which is the bottom margin on `.footer`. Once you open the software keyboard, the bottom margin stays, leaving a gap between input and keyboard. This has eventually be solved using javascript and checking for the keyboard being open.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -223,7 +223,9 @@ input[type=text], input[type=password], #sendMessage {
     overflow-y: auto;
     overflow-x: hidden;
     padding-top: 35px; /* topbar */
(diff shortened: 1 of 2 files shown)


bl1nk:
@lorenzhs The notch is handled with `viewport-fit: auto` already, which is the default. I created this to also handle the home-indicator on the bottom of the device.

PR #1030, opened 2018-10-29
Fix color issue in chan topic.

Daiyousei:
When the chan topic have some color definied, the full topic view print each color on multiple lines.

image

Now : 

image


lorenzhs:
Good catch, thanks!

PR #1040, opened 2018-12-16
[WIP] Implement support for TOTP (WeeChat >= 2.4)

lorenzhs:
TODO: proper UI, documentation, think about how to handle reconnects

Connecting works.

This builds on https://github.com/weechat/weechat/commit/738535e7f628d9b4a26cccc8b598e07e63ac1123 and https://github.com/weechat/weechat/commit/d0ea801724707c50517651955d6659e45d236f14

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,9 +100,9 @@
             <div class="panel-body">
                <form class="form-signin" role="form">
                 <div class="form-group">
-                  <label class="control-label" for="host">WeeChat relay hostname and port number</label>
                   <div class="input-group">
                     <div class="row no-gutter">
+                      <label class="control-label" for="host">WeeChat relay hostname and port number</label>
                       <div class="col-sm-9">
                         <input type="text" class="form-control favorite-font" id="host" ng-model="settings.host" placeholder="Address" autocapitalize="off">
                       </div>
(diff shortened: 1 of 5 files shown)


index.html
```diff
@@ -100,21 +100,28 @@ <h4 class="panel-title">
...
                         <input type="text" class="form-control favorite-font" id="port" ng-model="settings.port" placeholder="Port">
                       </div>
                     </div>
+                    <div class="row no-gutter">
+                      <label class="control-label" for="password">WeeChat relay password and TOTP token</label>
+                      <div class="col-sm-9">
+                        <input type="password" class="form-control favorite-font" id="password" ng-model="password" placeholder="Password">
+                      </div>
+                      <div class="col-sm-3">
+                        <input type="text" class="form-control favorite-font totp" id="port" ng-model="totp_token" placeholder="Token">
```
stgraber: shouldn't the `id` here be `totp`?

js/weechat.js
```diff
@@ -647,6 +648,10 @@
         if (params.password !== null) {
             keys.push('password=' + params.password);
         }
+        if (!!params.totp) {
+            keys.push('totp=' + params.totp);
+        }
+        console.log(keys);
```
stgraber: Presumably debugging that should go away before the branch is ready?

index.html
lorenzhs: ahem, yes of course.

js/weechat.js
lorenzhs: certainly, this is more of a proof of concept than a pull request at this stage :)

lorenzhs:
I dislike having the TOTP input field being shown by default because it looks like you need to enter something, even if you're not using 2FA. So the UI definitely needs to change for this to be merged. Maybe a "Use 2FA (TOTP)" checkbox that then shows the field.

Unfortunately, WeeChat's authentication flow doesn't allow us to validate the password before requesting the TOTP code. There's no way for Glowing Bear to know whether one is required. I dislike this choice, but we'll have to build around it I guess.

stgraber:
@lorenzhs so I take it that connecting without the TOTP token also doesn't result in an error indicating that TOTP is required? (so can't attempt connection with just password then prompt for TOTP and re-connect as needed)

lorenzhs:
@stgraber yeah, when I checked in December, it just closed the connection, with no way to tell what went wrong. I don't think that's changed.

AStove:
Sorry I didn't know this PR existed. The same feature implemented in #1097

Issue #1011, opened 2018-06-05
hostmasks should not be clickable like email addresses


rain-1:
In the buffer any time you see a hostmask, e.g.

```
--> foobar (foo@127.0.0.1) has joined #glowing-bear
```

the `foo@127.0.0.1` is a clickable link to `mailto:foo@127.0.0.1`.

These aren't email addresses and clicking on them by accident happens sometimes, which opens up an email client and it's a bit of an annoyance.

It seems to be coming from this, I think
https://github.com/angular/angular.js/blob/master/src/ngSanitize/filter/linky.js#L131

which doesn't provide any option to not insert mailto. So perhaps we could make a copy of that file and modify it.

rain-1:
I removed all mentions of 'linky' from index.html and added

```
    return function(text) {
        if (!text) {
            return text;
        }

        var urlRegex = /((https?:&#x2F;&#x2F;|www\.)[^\s]+)/gmi;
        var substitute = '<a href="$1" target="about:blank">$1</a>'
        text = text.replace(urlRegex, substitute);
```

in weechat.filter('irclinky' js/filters.js

This works but is it ok that the regex needs &#x2F; ?

rain-1:
This solution is not perfect yet, it has issues with links that have an anchor bit at the end.

rain-1:
```
diff --git a/index.html b/index.html
index 896e178..098a74a 100644
--- a/index.html
+++ b/index.html
@@ -267,7 +267,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
         </div>
 
         <div class="title"  ng-click="showModal('topicModal')">
-          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky')"></span>
+          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | DOMfilter:'irclinky')"></span>
           <span class="mobile" ng-class="{'active': activeBuffer().active, 'channel': activeBuffer().type === 'channel', 'channel_hash': activeBuffer().prefix === '#', 'channel_plus': activeBuffer().prefix === '+', 'channel_ampersand': activeBuffer().prefix === '&'}">{{ activeBuffer().trimmedName || activeBuffer().fullName }}</span>
         </div>
 
[...]

PR #1045, opened 2019-04-13
Don't linkify hostnames in join/part messages.

rscullin:
This adds a conditional `linky` filter, and won't be called for join/part hostnames. Accidentally clicking a `mailto:` link when someone joins a channel on my phone got fairly frustrating, so this seems to be a reasonable solution. This should resolve issue #1011.

 I'm looking for the `cof-chat_host` class, rather than looking for an `irc_join`/etc tag, as the class method lets things like URLs in quit messages get linkified, while not linkifying hostnames. I'll re-work this if there's a better way.

Unconditional linky:
image

Conditional linky:
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -337,7 +337,6 @@ npm run build-electron-{windows, darwin, linux}</pre>
            --><td class="message"><!--
              --><div ng-repeat="metadata in ::bufferline.metadata" plugin data="::metadata"></div><!--
              --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | conditionalLinkify:part.classes.includes('cof-chat_host') | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' | DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
-             <!-- <pre>{{bufferline | json}}</pre> -->
               </td>
             </tr>
             <tr class="readmarker" ng-if="activeBuffer().lastSeen==$index">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -336,7 +336,8 @@ <h4 class="panel-title">
...
            --><td class="message"><!--
              --><div ng-repeat="metadata in ::bufferline.metadata" plugin data="::metadata"></div><!--
-             --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' | DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
+             --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | conditionalLinkify:part.classes.includes('cof-chat_host') | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' | DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
+             <!-- <pre>{{bufferline | json}}</pre> -->
```
lorenzhs: can you amend the commit to remove this please?

lorenzhs:
Thanks, looks nice! I just configured WeeChat not to show the hostnames at all ;)

index.html
rscullin: Oops! Yeah, I'll fix that.

lorenzhs:
Brilliant, thanks very much!

Issue #1052, opened 2019-07-01
Code highlighting?


spapas:
Could we add a switch to enable code highlighting using backticks (``) ? For example somebody could write 

> Here's my code snippet: \`this.is.a.code = snippet\`

and the \`this.is.a.code = snippet\` would be highlighted. From what I see in other clients (like irccloud) it just puts it in a &lt;pre&gt; html tag (with a monospaced font and a border). 

I know that there are various javascript libraries for code highlighting (f.e https://highlightjs.org/) but i'm not sure these will be actually useful in this case since they need to know the language in order to format it properly and inline snippets are small and most users wouldn't want to actually define the language for their snippet. 

This can also be enabled using a switch in settings like the latex math rendering .

lorenzhs:
Since highlightjs' homepage states "automatic language detection" as a feature, I guess this could work. This would definitely be cool to have!

spapas:
Hello @lorenzhs I tried using highlight.js but I don't like the results that much :| The language detection isn't working that good and it is actually recommended to specify it for small snippets: https://github.com/highlightjs/highlight.js/issues/1830 

Well this is something expected; how could anybody know that print("Hello"); is Javascript or C ?

In any case, I also had various problems trying to style the highlighted snippets for each line. The default styles for highlightjs are to be used with blocks and not inline (as we'd need in this case).

As a solution I recommend my initial proposal of just adding a code/pre around the backticked snippets. I tried it a bit and it worked fine.

lorenzhs:
Yeah, that sounds reasonable and is a lot less invasive. Good idea!

PR #1053, opened 2019-07-03
Add a simple codify filter

spapas:
This should implement #1052

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -239,8 +239,7 @@ weechat.filter('prefixlimit', function() {
 
 weechat.filter('codify', function() {
     return function(text) {
-        var re = /(`[^``]+`)/g;
-
+        var re = /(`.+?`)/g;
         return text.replace(re, function(z) {
             var rr = '<code>' + z.slice(1, z.length-1) + '</code>';
             return rr;
--- a/js/filters.js
+++ b/js/filters.js
@@ -237,4 +237,15 @@ weechat.filter('prefixlimit', function() {
     };
 });
 
+weechat.filter('codify', function() {
+    return function(text) {
+        var re = /(`[^``]+`)/g
+
+        return text.replace(re, function(z) {
+            var rr = '<code>' + z.slice(1, z.length-1) + '</code>';
+            return rr;
+        })
+    };
+});
+
 })();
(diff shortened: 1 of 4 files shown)


lorenzhs:
What would you say should be the expected behaviour for snippets containing URLs? I think in ``, codify should take precedence over the URL linkification (`irclinky`) and thus be applied *before* it. GitHub also does it that way.

Other than that, I really like it! Thanks a lot!

spapas:
Hello @lorenzhs after some research I found out that the `DOMfilter` filter splits the code containing links to multiple strings thus if a string like "foo \`http://www.bar.baz bar\`" is input the `codify` filter will get two strings as an input: 

foo \`

and 

http://www.bar.baz bar\`

thus it won't work properly :(

A solution I was able to found was to call codify directly right after the `conditionalCodify` without `DOMFilter` i.e the filters will be like

``::part.text | conditionalLinkify:part.classes.includes('cof-chat_host') | codify | DOMfilter:'irclinky'  | ``

Is it ok to proceed like this ?

lorenzhs:
Sorry, I meant before `conditionalLinkify` :\ `irclinky` only does linkification of irc channel names.

Just wrapping things in a `<code>` tag without escaping is no good because it is vulnerable to cross-site scripting. I forgot whether angular handles this, sorry.

spapas:
Hello @lorenzhs maybe i didn't make myself clear so I've pushed my commit with the fix to take a look at. 

There shouldn't be any vulnerabilities to cross-site scripting with the `codify` filter, it just adds the `<code>` tags where needed and passes everything to the other filters so every conversion of non-safe code should still be performed. 

Please take a look at it thank you!

js/filters.js
```diff
@@ -237,4 +237,14 @@ weechat.filter('prefixlimit', function() {
     };
 });
 
+weechat.filter('codify', function() {
+    return function(text) {
+        var re = /(`.+?`)/g;
```
lorenzhs: Why `.+?`? Isn't that equivalent to `.*`?

lorenzhs:
Yeah I think since you don't have any `$sce.trustAsHtml` thingies, it should be okay.

@torhve can you have a look? Is this XSS-safe?

torhve:
I don't have the bandwidth to do full testing, but I think this makes sense, I like this change

lorenzhs:
Thanks a lot, @spapas!

PR #1057, opened 2019-08-18
Add settings option to toggle codify filter

mcpower:
This PR implements a settings option to toggle the `codify` filter on and off, similar to the settings for the `emojify` and the `latexmath` filters. The option defaults to on, which retains the default behaviour before this PR.

I personally prefer to keep the `codify` filter off, as it visually modifies the content of the message (removing backticks). The `codify` filter also causes some unexpected behaviour when viewing text with backticks, such as <code>&#96;&#96;fancy'' &#96;&#96;quotes''</code> or <code>http://google.com/`weird`</code> - that last one is probably a bug.


mcpower:
Superceded by #1059, closing.

mcpower:
If anyone stumbles upon this PR and is interested in turning off the codify filter, here's some CSS which effectively turns the filter off:

```css
.message .hidden-bracket {
  position: static;
}

code {
  padding: 0px;
  font-size: 100%;
  border: none;
  color: inherit;
  background-color: inherit;
}
```

I'd personally recommend removing the `color` and the `background-color` from this CSS so you still get some coloured styling on code snippets :)

lorenzhs:
This would be a great addition to the CSS Customizations wiki page at https://github.com/glowing-bear/glowing-bear/wiki/CSS-Customizations

mcpower:
I've added the code snippet to the wiki page. Thanks for the link - I never knew that page existed!

PR #1059, opened 2019-08-19
Improve codify filter

### LINKED ISSUE #1058: Backticks in URLs causes problems with codify filter
[The `text.replace` in the `codify` filter](https://github.com/glowing-bear/glowing-bear/blob/6ab4ef7d88668bfb7a985fef48b3923787f4d95b/js/filters.js#L243-L246) replaces backticks everywhere in the text, including in HTML tags - most notably `<a>`s generated by the `linkify` filter.

For example, the message

    http://google.com/`weird` one http://google.com/`weird` two http://google.com/`weird` three

is rendered like

screenshot

due to there being two backticks in the `href` attribute of the anchors, and the message

    `http://google.com/` one `http://google.com/` two `http://google.com/` three

is rendered like

screenshot

due to there being one backtick in the `href` attribute of the anchors.

One way to fix this is to walk the DOM in the `codify` filter and only apply the backtick replacement on top-level text nodes... That seems a bit complicated for a simple formatting filter - is there a simpler way?

lorenzhs:
- Only match after whitespace or at the start of the line
- Match triple backticks, too
- Document the new, more complicated regex :)
- Theme the code elements in the dark and black themes
- Add tests to verify that these changes work

Fixes #1058
Supersedes #1057 because it catches all of the cases described therein :)

### DIFF
--- a/css/themes/dark.css
+++ b/css/themes/dark.css
@@ -2114,6 +2114,11 @@ button.close:hover {
     font-weight: bold;
 }
 
+code {
+  background-color: #444;
+  color: #fff;
+}
+
 /* */
 /* Mobile layout */
 /* */
--- a/index.html
+++ b/index.html
@@ -336,7 +336,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
               <td class="prefix"><span ng-class="::{'repeated-prefix': bufferline.prefixtext==bufferlines[$index-1].prefixtext}"><a ng-click="addMention(bufferline)"><span class="hidden-bracket" ng-if="::(bufferline.showHiddenBrackets)">&lt;</span><span ng-repeat="part in ::bufferline.prefix" ng-class="::part.classes" ng-bind="::part.text|prefixlimit:25"></span><span class="hidden-bracket" ng-if="::(bufferline.showHiddenBrackets)">&gt;</span></a></span></td><!--
            --><td class="message"><!--
              --><div ng-repeat="metadata in ::bufferline.metadata" plugin data="::metadata"></div><!--
-             --><span ng-repeat="part in ::bufferline.content" class="text" ng-class="::part.classes.concat(['line-' + part.$$hashKey.replace(':','_')])" ng-bind-html="::part.text | conditionalLinkify:part.classes.includes('cof-chat_host') | codify | DOMfilter:'irclinky' | DOMfilter:'emojify':settings.enableJSEmoji | DOMfilter:'inlinecolour' |  DOMfilter:'latexmath':('.line-' + part.$$hashKey.replace(':','_')):settings.enableMathjax"></span>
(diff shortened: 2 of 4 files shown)


mcpower:
Awesome - this should fix #1058 cleanly and handle most bad codify cases! :+1:

>Supersedes #1057 because it catches all of the cases described therein :)

I'm not totally convinced about this. I can't write up a full explanation as I'm short on time (I may post a follow up comment later), but here are my vague thoughts:

- codify, like latexmath, gives special meaning to some characters for formatting purposes - in a medium which is meant to be plain text (barring widely supported colour codes / format codes etc.)
- messages rendered with said filters will be formatted noticeably different compared to the weechat CLI (c.f. inlinecolour or plugins, which only add in unobtrusive extra things to the message)
- there will *always* be cases where the characters will be "misinterpreted" as they are meant to be interpreted as plain text - no matter how robust your detection is
- latexmath can be disabled, codify cannot be

lorenzhs:
Latexmath is a setting because a) it's a multi-megabyte download and b) it can slow things down a lot. Neither of these applies to codify. Also, copying the message yields the exact original from WeeChat, so at least that is maintained. We're trying to be a modern web frontend to IRC here that makes sane choices for most users. That means making choices that some users may disagree with. And while each of these issues in isolation would be easy to implement, in sum, they add up to a big maintenance burden.

And seriously, what's the harm of displaying something in monospace font? Everyone who's ever used Markdown is going to know exactly what happened and be somewhat amused, but very unlikely to be confused. I just don't see the point of adding a setting.

mcpower:
> Everyone who's ever used Markdown is going to know exactly what happened and be somewhat amused, but very unlikely to be confused.

Good point. This PR fixes all of the "confusing" cases of codify, so I agree that it's clear to the user what happened when monospace font appears out of nowhere. Keen users can theme the `.hidden-bracket` to be visible if they are bothered by the missing backticks. I'll close #1057 now - apologies for the hassle.

With this PR, codifying `http://google.com/` doesn't work as expected, creating a link with the text

    `">http://google.com/`

linking to

   http://google.com/%3C/code%3E%3Cspan%20class=

similar to the behaviour before this PR. I think requiring whitespace or the end of the line *after* the final backtick should fully fix #1058, which comes from the regex matching the backtick inside the `<a href="...">` tag.

lorenzhs:
Oh, good catch on the broken linkification of URLs! We need to make linkify use URLfilter to fix that, I guess. I'll have a look.

mcpower:
> this causes the quoted link to not be codified, but I think that's okay.

Agreed - this is probably the best way of dealing with links (and potentially other HTML elements in the future!). DOMfilter should prevent the filter from replacing backticks within HTML tags, so this looks pretty hard to break.

I've done a bit of testing and it seems like everything is working as intended - LGTM!

lorenzhs:
What DOMfilter does is apply a filter to the content of every HTML element so that it doesn't further mess up things. This makes it safe to apply filters to nested content. Otherwise, the filter logic would have to be terribly complicated and failure-prone, not to mention a huge potential for XSS vulnerabilities. So it does handle other elements than links already :)

PR #1069, opened 2019-11-20
Added option to always show unread count badge in mobile

Fmstrat:
Created a setting only viewable in mobile that keeps the unread count badge viewable. Currently it does not display in the top-bar when in mobile view, which is a usability issue on a desktop machine when docking to half a monitor.

gb

### DIFF
--- a/index.html
+++ b/index.html
@@ -486,7 +486,16 @@ npm run build-electron-{windows, darwin, linux}</pre>
                   </div>
                 </form>
               </li>
-              <li>
+              <li class="mobile">
+                <form class="form-inline" role="form">
+                  <div class="checkbox">
+                    <label>
+                      <input type="checkbox" ng-model="settings.alwaysunreadbadge">
+                      Always show unread count in favicon
+                    </label>
+                  </div>
+                </form>
+              </li>              <li>
                 <form class="form-inline" role="form">
                   <div class="checkbox">
                     <label>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -259,8 +259,8 @@ <h4 class="panel-title">
           <a href="#" ng-click="toggleSidebar()">
             <img alt="brand" src="assets/img/favicon.png" title="Connected to {{ settings.host }}:{{ settings.port}}">
           </a>
-          <span class="badge" ng-show="unread > 0">{{unread}}</span>
-          <span class="badge danger" ng-show="notifications > 0">{{notifications}}</span>
+          <span class="badge" ng-class="{'badge-show': showUnreadBadge}" ng-show="unread > 0">{{unread}}</span>
+          <span class="badge danger" ng-class="{'badge-show': showUnreadBadge}" ng-show="notifications > 0">{{notifications}}</span>
```
lorenzhs: you can use `settings.alwaysunreadbadge` directly here and save yourself the callback further down

lorenzhs:
Great idea! A few things though:
- That's not the favicon :) Mobile browsers tend to not show them at all; they're the little icon in the tabs of desktop browsers.
- Tapping the badge should open the buffer list, not the topic. It should probably get an `ng-click="toggleSidebar()"` (hopefully the tap target is big enough)
- What was the idea behind #581, which purposefully hid the badge on mobile? I think we still had the topic in the top bar on mobile at that point, so the idea was probably to save space. But the top bar now contains the channel name, so that reasoning is void.
- I don't see a reason why this should be a setting and not just default behaviour.
- In that case, reverting #581 would be the much simpler change :)

Fmstrat:
Totally down for that, just didn't want to step on toes. I will remove.

lorenzhs:
Thanks, this is a nice feature. After playing around with it for a few minutes, I like it a lot. Sometimes it takes a fresh pair of eyes to notice that things don't make sense as they are :)

PR #1075, opened 2019-11-22
Update node dependencies

lorenzhs:


lorenzhs:
@baldurmen none of these are strictly needed, you can stay on older versions for packaging. But there's not much of a point of staying on older versions (which most of the time aren't getting bugfixes/...) if you don't need to, which is the idea of this pull reuqest

baldurmen:
Thanks for the heads-up.

PR #1076, opened 2019-11-22
Stop repo modification when building Electron apps

Fmstrat:
Currently building an Electron version risks including rogue `index.html` and `3rdparty` files in a commit. This PR helps solve this and is a precursor to supporting `electron-builder` as part of GitHub Actions.

- Creates a `build` directory for building electron apps
- Creates a `dist` directory where builds end up upon completion
- Includes `build-electron-{linux|darwin|windows}-archive` scripts to create `zip` and `tgz` archives as options
- Includes the Windows x64, favicon, and asar fix from https://github.com/glowing-bear/glowing-bear/pull/973 to support non-error Window's builds


lorenzhs:
This is a good idea, but what's with the 1MB favicon.ico?

lorenzhs:
Closing since we removed electron

PR #1077, opened 2019-11-28
Fix highlights notify_private buffers with where notify level is 0

### LINKED ISSUE #1031: Stop IRC Server Buffers From Generating Highlight Notification
I had mentioned this on IRC the other day, but would like to know if anyone else has an idea on how to fix this as it does become rather annoying when you an oper and every message in the IRC server buffer generates a highlight notification/message. 

Thanks!

Techman:
Fixes #1031

### DIFF
--- a/js/handlers.js
+++ b/js/handlers.js
@@ -168,7 +168,7 @@ weechat.factory('handlers', ['$rootScope', '$log', 'models', 'plugins', 'notific
                     $rootScope.$emit('notificationChanged');
                 }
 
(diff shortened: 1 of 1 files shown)


lorenzhs:
This can be simplified as
```js
if (buffer.notify !== 0 && (message.highlight || _.contains(message.tags, 'notify_private')))
```
which would make it clearer what is happening.

I can't really test this because I never get notifications for things going on in server buffers, but if @Fmstrat and you confirm that it solves the problem, I guess that's fine. You can try it out at https://latest.glowing-bear.org/pull/1077/

lorenzhs:
Let's see if it works, I guess ;)

PR #1081, opened 2019-12-08
Don't show a TLS warning when loading locally

baldurmen:
I think that's the second time ever I write JS. I tested the patch and it seems to work, but I'm happy to make changes if that's not the right way to do it.


lorenzhs:
Good idea, thanks. I think there are a few things that local pages aren't allowed to do, for example, none of the keybindings work for me when loading GB from a `file://` URL (Firefox 71). But that doesn't mean we should warn about something unrelated ;)

PR #1085, opened 2019-12-08
Update http-server to mitigate INVALID_REDIRECT bug

AStove:
Had this INVALID_REDIRECT bug when I opened the app locally. This upgrade of http-server fixes it.

https://stackoverflow.com/questions/56364464/http-server-with-localhost3000-gives-err-invalid-redirect


lorenzhs:
WTF, JS-verse, WTF. Thanks.

PR #1087, opened 2019-12-09
Correctly interpret value for addSpace

### LINKED ISSUE #1086: Autocompleting in the middle of a line eats the space after the nickname
When nick completing in the middle of a line it eats the space after the nick

`foo n bar` get autocompleted to `foo nicknamebar`

AStove:
This value is not a boolean. Either do it like this, or parse the values correctly when getting the values. This is one in https://github.com/glowing-bear/glowing-bear/blob/master/js/handlers.js#L16

Not sure which makes more sense. But if you start interpreting booleans, you'd have to do it for all types, and you'd have to go over all the code to check every spot where such values are used.

fixes #1086

### DIFF
--- a/js/irc-utils.js
+++ b/js/irc-utils.js
@@ -119,7 +119,7 @@ IrcUtils.service('IrcUtils', [function() {
             suf = ':';
         }
         // addSpace defaults to true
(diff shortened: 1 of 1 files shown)


lorenzhs:
Ugh, good catch. I think this is one for @torhve

PR #1089, opened 2019-12-11
Remove nicklist when leaving channel and bring it back when needed

### LINKED ISSUE #1088: Nicklist stays hidden when rejoining a channel
If you rejoin a channel while you have the buffer of said channel open, the nicklist stays hidden.

To reproduce:
1. Join a channel that is +r without being identified, /join <span>#</span>test, buffer opens but there is no nicklist because you couldn't join (normal)
2. Identify so you can now join
3. While in buffer <span>#</span>test, do /join <span>#</span>test

Additionally when /leaving a channel the nicklist is supposed to close.

AStove:
Fixes #1088 

Please feel free to correct my code and style, etc. I tried to touch as little as possible but had to go further than I wanted it to. Nonetheless I think it's a pretty clean solution and it actually brings a nice feature, The nicklist should really dissapear when you are not joined in a channel.

Tested well in desktop chrome but did not have a chance to do it on mobile or electron.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -149,6 +149,9 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }, false);
     }
 
+    $rootScope.$on('nickListChanged', function() {
+            $scope.updateShowNicklist();
+    });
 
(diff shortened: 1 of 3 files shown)


lorenzhs:
Thanks! I agree that hiding the nicklist on buffers you haven't joined is a good idea.

PR #1090, opened 2019-12-15
Allow specifying custom websocket path in host field

### LINKED ISSUE #1008: Allow specifying custom websocket path
Currently, it seems to be [hard-coded](https://github.com/glowing-bear/glowing-bear/blob/84018d4bec3d17cd9edafcd55843054959b9d992/js/connection.js#L30).

AStove:
- Makes the host field into a multi purpose field as suggested in #1008.
- Disables the port input if it's already specifed in the host field.
- Preserves saved settings, shouldn't break on update
- Added a paragraph in the usage instructions
- Change how the bg color of disabled inputs, it was too white. Not sure if this has in influence on custom styling and such.
- Tested with strings:
```
192.168.0.1:8000
192.168.0.1:8000/weechat2
192.168.0.1
[2001:0db8:85a3:0000:0000:8a2e:0370:7334]
2001:0db8:85a3:0000:0000:8a2e:0370:7334 (does not work, intentionally)
2001:0db8:85a3:0000:0000:8a2e:0370:7334/weechat2  (does not work, intentionally)
[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8000
[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:8000/weechat2
test.com
test.com:8000
test.com:8000/weechat2
```

resolves #1008

(I'm new to this. There's some extra commits in my pull request that were due to me updating my own downstream branch. I'm not sure if those commits need to be deleted or not.)

### DIFF
--- a/index.html
+++ b/index.html
@@ -111,6 +111,8 @@
                       </div>
                     </div>
                   </div>
+                  <label class="control-label" for="path">WebSocket path</label>
+                  <input type="text" class="form-control favorite-font" id="path" ng-model="settings.path" placeholder="WebSocket path">
                   <label class="control-label" for="password">WeeChat relay password</label>
                   <input type="password" class="form-control favorite-font" id="password" ng-model="password" placeholder="Password">
                   <div class="alert alert-danger" ng-show="passwordError" ng-cloak>
--- a/index.html
+++ b/index.html
@@ -207,10 +207,41 @@ chown -R <strong>username</strong>:<strong>username</strong> ~<strong>username</
               <p>
               Helpful trigger to automatically repin a buffer (in this instance, <var>irc.freenode.#weechat</var>): <pre><code>/trigger add autopin signal "buffer_opened" "${buffer[${tg_signal_data}].full_name} =~ <var>irc.freenode.#weechat</var>" "" "/command -buffer ${buffer[${tg_signal_data}].full_name} * /buffer set localvar_set_pinned true"</code></pre>
               </p>
-              <h3>Setting path</h3>
-              <p>
-                The path is by default 'weechat'. In case a proxy is used the path can be changed by entering it in the host field. For example <code>your.domain.com:8000/otherpath</code>.
+              <h3>Setting a custom path</h3>
+              <p> 
+                To connect to the weechat relay service we connect using a URL. A typical URL consists of 4 parts. <code>{scheme}://{host}:{port}/{path}</code>. The path can be changed by enterying the relay's full URL (except the scheme).
+              </p> 
+                <ul>
+                  <li><b>scheme</b>: The scheme must never be input. The scheme is "ws" if TLS isn't used and it is "wss" if TLS is used.</li>
+                  <li><b>host</b>: Can be an IPv4, IPv6 or a FQDN. IPv6 addresses must be wrapped in square brackets.</li>
+                  <li><b>port</b>: can be specified in the host field or the seperate port field. However if the path is specified in the host field the port must also be specified.</li>
+                  <li><b>path</b>: by defautl this is "weechat". In case <a href="https://github.com/glowing-bear/glowing-bear/wiki/Proxying-WeeChat-relay-with-a-web-server">a proxy</a> is used the path can be changed by entering it in the host field.</li>
+                </ul>
+              <p> 
+                Examples of correct input for the host field are:
               </p>
+                <ul>
+                  <li>192.168.0.1</li>
+                  <li>192.168.0.1:8000</li>
+                  <li>192.168.0.1:8000/weechat2</li>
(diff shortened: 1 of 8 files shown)


AStove:
I changed back the CSS edit because this only applies to one theme. This is more an issue with the themes than anything, added an issue #1091.

index.html
```diff
@@ -207,6 +207,10 @@ <h3>Pinning buffers</h3>
               <p>
               Helpful trigger to automatically repin a buffer (in this instance, <var>irc.freenode.#weechat</var>): <pre><code>/trigger add autopin signal "buffer_opened" "${buffer[${tg_signal_data}].full_name} =~ <var>irc.freenode.#weechat</var>" "" "/command -buffer ${buffer[${tg_signal_data}].full_name} * /buffer set localvar_set_pinned true"</code></pre>
               </p>
+              <h3>Setting path</h3>
+              <p>
+                The path is by default 'weechat'. In case a proxy is used the path can be changed by entering it in the host field. For example <code>your.domain.com:8000/otherpath</code>.
```
lorenzhs: This should be made clearer, an example isn't a good explanation. It should:
- have a link to the reverse proxy wiki page
- the path can be changed by entering *the relay's full URL* it in the host field
- explain that the URL consists of host colon port slash path

js/glowingbear.js
```diff
@@ -656,6 +657,39 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+            $rootScope.portDisabled = true;
+        }
+        //host:port/path
+        else if((parts = regexHostPortPath.exec(settings.host)) !== null)
+        {
+            settings.hostOnly = parts[1];
+            settings.port = parts[2];
+            settings.path = parts[3];
+            $rootScope.portDisabled = true;
+        }
```
lorenzhs: what if nothing matches and the input is garbage? Maybe we should make the input have a red border to indicate the error or something. I'm sure bootstrap has something for that.

js/glowingbear.js
```diff
@@ -656,6 +657,39 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+
+        var parts;
+
+        //host
+        var regexHost = /^([^:\/]*|\[.*\])$/;
+        var regexHostPort = /^([^:]*|\[.*\]):(\d+)$/;
+        var regexHostPortPath = /^([^:]*|\[.*\]):(\d*)\/(.+)$/;
+        if((parts = regexHost.exec(settings.host)) !== null)
+        {
+            settings.hostOnly = parts[1];
```
lorenzhs: `hostOnly` is not a great name; should probably rename the old `host` to `hostField` (and bind the input to that) and use `host` for the new one. That would require handling the case where `host` is set but `hostField` is not upon page load, but I think the increased clarity is worth it.

index.html
AStove: I've expanded this section a bit considering your suggestions.

lorenzhs: "a bit" :D

js/glowingbear.js
AStove: Added validation for the input field and also disabled the connect button. Added a red border for all three styles.

js/glowingbear.js
AStove: I agree and I had thought about it but it seemed too invasive to touch the saved settings. Now I changed the names to the more logical ones like you suggested but had to add a bit of code for when users upgrade or they'll lose their settings. This extra bit of code should probably dissapear in the far future (as if).

AStove:
I've added an entry in the themes for darker or lighter background color for the disabled input fields. For the port field now.

resolves #1091

js/glowingbear.js
```diff
@@ -66,6 +66,12 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     });
     $scope.settings = settings;
 
+    //For upgrade reasons because we changed the name of host to hostField
+    //check if the value might still be in the host key instead of the hostField key
+    if (!settings.hostFieldv && settings.host) {
```
lorenzhs: there's a stray `v` here in `hostFieldv`

lorenzhs: does this need to trigger `parseHost`?

lorenzhs:
This is fantastic! Thanks so much, @AStove!

js/glowingbear.js
AStove: I don't think it does because it's already triggered at init and this is in the constructor before init. Thanks for fixing the typo and the merge.

lorenzhs: Yeah it worked fine in testing

PR #1095, opened 2019-12-16
Hide sidebar for all widths according to state and correctly apply state

### LINKED ISSUE #1094: Sidebar overlaps the bufferlines after dragging tab
When draging the tab in chrome and detaching the tab from the current window, and sticking it to a new window there is a chance that the sidebar will overlap the bufferlines.

The sidebar is always displayed correctly however sometimes the bufferlines don't get the `sidebar-state=visible`. This prevents the margin-left from being applies so it's under the sidebar. The sidebar also has a different width when this happens because the `#sidebar[data-state=visible]` is not applied. See screenshots for what I mean.

The reason I think this happens is it resizes the window to below the mobile size limit. If you then add it to a full size window again it resizes again. It doesn't always seem to reproduce. If you have no success in reproducing try having a different buffer open 

Bug presents itself:

Correct for comparison:

AStove:
Fixes #1094 

Change in the CSS is because the sidebar should always be hidden if the state is hidden, not only for narrow windows.

Change wasMobileUi check when showing sidebar because this was sometimes incorrect when rapidly resizing windows like what happens when dragging a tab in chrome.

Doesn't seem to have a performance impact on desktop, did not check on mobile. I don't think the performance impact can be that much from setting a few of attributes.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -263,6 +263,11 @@ input[type=text], input[type=password], #sendMessage {
 #sidebar.ng-hide {
   width: 0;
 }
+    
+#sidebar[data-state=hidden] {
+    transform: translate(-200px,0);
+    -webkit-transform: translate(-200px,0);
+}
 
 #nicklist {
     position: fixed;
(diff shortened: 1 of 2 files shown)


lorenzhs:
FTR, `wasMobileUi` was introduced in #241, but the issue mentioned there doesn't seem to apply any more. I tested this on iOS and it works great. Thanks!

PR #1096, opened 2019-12-18
Fix atob error, add url params

### LINKED ISSUE #1005: Can't pass credentials from URL
Seems like code merged in #644 no longer works, as now `window.location.hash` seems to be prepended with `#!/`. This should be really easily fixable by changing `window.location.hash.substring(1)` to `window.location.hash.substring(3)`.

AStove:
fixes #1005
This started out as just the issue fix but found it a weird way to do it so I tried to improve it. It now also no longer gives the atob error in the console.

- Remove /#! from the url this is foreign for most web stuff and confusing. Old URLs should keep working by redirecting to the /.
- Add url parameters
- Add documentation for the url paramters.

### DIFF
--- a/js/glowingbear.js
+++ b/js/glowingbear.js
@@ -8,9 +8,12 @@ document.addEventListener("deviceready", function () {
     }
 }, false);
 
-var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'bufferResume', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch'], ['$compileProvider', function($compileProvider) {
+var weechat = angular.module('weechat', ['ngRoute', 'localStorage', 'weechatModels', 'bufferResume', 'plugins', 'IrcUtils', 'ngSanitize', 'ngWebsockets', 'ngTouch'], ['$compileProvider', '$locationProvider', function($compileProvider, $locationProvider) {
     // hacky way to be able to find out if we're in debug mode
     weechat.compileProvider = $compileProvider;
+
+    //remove hashbang from url
+    $locationProvider.html5Mode(true).hashPrefix('');
 }]);
 weechat.config(['$compileProvider', function ($compileProvider) {
     // hack to determine whether we're executing the tests
@@ -19,8 +22,8 @@ weechat.config(['$compileProvider', function ($compileProvider) {
     }
 }]);
 
-weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout', '$log', 'models', 'bufferResume', 'connection', 'notifications', 'utils', 'settings',
-    function ($rootScope, $scope, $store, $timeout, $log, models, bufferResume, connection, notifications, utils, settings)
+weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout','$location', '$log', 'models', 'bufferResume', 'connection', 'notifications', 'utils', 'settings',
+    function ($rootScope, $scope, $store, $timeout, $location, $log, models, bufferResume, connection, notifications, utils, settings)
 {
 
     window.openBuffer = function(channel) {
@@ -699,6 +702,7 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     };
 
     $scope.connect = function() {
+
         notifications.requestNotificationPermission();
         $rootScope.sslError = false;
         $rootScope.securityError = false;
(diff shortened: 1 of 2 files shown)


lorenzhs:
Hm, https://latest.glowing-bear.org/pull/1096/?host=example.com doesn't work for me, neither in Firefox nor in Chrome. Both redirect to https://latest.glowing-bear.org/pull/1096/?host=example.com#!/ (with the trailing `#!`. Any idea why?

AStove:
Ah, yes. Because in the HTML I've added `<base href="/">` This now takes the javascript files from / and not from /pull/1096. (I did not know this existed.) So you are testing with all the .js files from the lastest build and not from the PR. I'll try to play around with it some more to see if this `<base>` tag is strictly nescessary.

lorenzhs:
Ugh. That's a breaking change (not just for latest, people might have deployed GB at any location they like), we can't do that, and definitely not for a feature that only a handful of people use

AStove:
Not if we can set the base dynamically on runtime.

AStove:
Ok so, sorry for the commit spam. I had some problems with clearing the cache because it kept redirecting.

- Removed a part where the hash would be used to autoconnect. I don't know what the purpose of that was since the autoconnect is already saved in storage.
- No more base
- Old links with #! should still work like before

Some links for testing, be sure to clear the cache like always:

- https://latest.glowing-bear.org/pull/1096/?host=example.com
- https://latest.glowing-bear.org/pull/1096/?host=example.com&port=666
- https://latest.glowing-bear.org/pull/1096/?host=example.com&port=666&autoconnect=true
- https://latest.glowing-bear.org/pull/1096/?path=testpath

lorenzhs:
Aaah uh sorry I didn't notice this before, but you're currently using GET parameters. These are included in the request to the server, meaning using it this way would send all connection details included in the URL to the server, which is an absolute no-go. Using the URL hash avoids this because the hash is never sent to the server.

AStove:
This time with hash parameters instead of get parameters

https://latest.glowing-bear.org/pull/1096/#host=example.com
https://latest.glowing-bear.org/pull/1096/#host=example.com&port=666
https://latest.glowing-bear.org/pull/1096/#host=example.com&port=666&autoconnect=true
https://latest.glowing-bear.org/pull/1096/#path=testpath
https://latest.glowing-bear.org/pull/1096/#!/ (does nothing, just to check if old bookmarks would break

js/glowingbear.js
```diff
@@ -687,7 +687,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         }
     };
 
+    
+    $scope.parseHash = function() {
+
+        //Fill in url parameters, they take precedence over the stored settings, but store them
+        var params = $location.$$hash.split('&').map(function(val) { return {key: val.split('=')[0], value: val.split('=')[1]}; });
```
lorenzhs: according to the linter there's a semicolon missing on this line in column 126 but tbh it's so dense I can't really parse it

js/glowingbear.js
```diff
@@ -687,7 +687,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+        var hostParam = params.find(function(p) { return p.key === 'host'; });
+        var portParam = params.find(function(p) { return p.key === 'port'; });
+        var pathParam = params.find(function(p) { return p.key === 'path'; });
+        var passwordParam = params.find(function(p) { return p.key === 'password'; });
+        var autoconnectParam = params.find(function(p) { return p.key === 'autoconnect'; });
+        if(hostParam) { $scope.settings.host = hostParam.value; $scope.settings.hostField = hostParam.value;}
+        if(portParam) { $scope.settings.port =  parseInt(portParam.value);}
+        if(pathParam) { $scope.settings.path = pathParam.value;}
```
lorenzhs: If `pathParam` is set, I think `settings.hostField` should be set to `host + ':' + port + '/' + path` so that the changed path is visible in the UI

js/glowingbear.js
lorenzhs: In JS, mappers can have side effects, so you can do
```js
var params;
$location.$$hash.split('&').map(function(val) {
    var segs = val.split('=');
    params[segs[0]] = segs[1];
});
```
and then you can simply use `params['host']` etc

js/glowingbear.js
```diff
@@ -687,7 +687,26 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
+        if(passwordParam) { $scope.settings.password = passwordParam.value;}
```
lorenzhs: The model for the password field is not `settings.password`, it's a password field in the scope which is saved to settings if the checkbox is set. I don't think it should just be saved to the settings if set in the hash.

AStove: Oh ok, this is not intentional, copy pase error.

js/glowingbear.js
AStove: I initially went for a find() with a lambda but this is not ES6.

AStove: Or even dot notation params.host, the linter suggested.

lorenzhs:
Thanks a lot! Rebased and merged.

PR #1097, opened 2019-12-18
Support TOTP

AStove:
-Add checkbox to use TOTP
-This checkbox unhides TOTP input field
-This checkbox disable autoconnect
-Contents of TOTP field is sent to weechat
-Added usage instructions

Thoughts on sanitizing the TOTP field? It's always a 6 digit int (with leading zero's), should we enforce that? With full validation making the inputfield red etc?

### DIFF
--- a/index.html
+++ b/index.html
@@ -120,7 +120,7 @@
                     </div>
                     <div class="col-sm-3" ng-Show="settings.useTotp">
                       <label class="control-label" for="totp">Token</label>
-                      <input type="text" class="form-control favorite-font" id="totp" ng-model="totp" autocomplete="off">
+                      <input type="text" class="form-control favorite-font" id="totp" ng-model="totp" ng-change="parseTotp()" ng-class="{'is-invalid': totpInvalid}" autocomplete="off">
                     </div>
                   </div>
                   <div class="alert alert-danger" ng-show="passwordError" ng-cloak>
@@ -152,7 +152,7 @@
                     </label>
                   </div>
                 </div>
-                <button class="btn btn-lg btn-primary" ng-disabled="hostInvalid" ng-click="connect()" ng-cloak>{{ connectbutton }}  <i ng-class="connectbuttonicon" class="glyphicon"></i></button>
+                <button class="btn btn-lg btn-primary" ng-disabled="hostInvalid || (totpInvalid && settings.useTotp)" ng-click="connect()" ng-cloak>{{ connectbutton }}  <i ng-class="connectbuttonicon" class="glyphicon"></i></button>
               </form>
             </div>
           </div>
--- a/index.html
+++ b/index.html
@@ -114,7 +114,7 @@
                   <label class="control-label" for="password">WeeChat relay password</label>
                   <input type="password" class="form-control favorite-font" id="password" ng-model="password" placeholder="Password">
                   <div class="alert alert-danger" ng-show="passwordError" ng-cloak>
-                    Error: wrong password
+                    Error: wrong password or one-time password
                   </div>
 
                   <div class="checkbox">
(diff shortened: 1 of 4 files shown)


js/connection.js
```diff
@@ -328,7 +330,8 @@ weechat.factory('connection',
     var attemptReconnect = function (bufferId, timeout) {
         $log.info('Attempting to reconnect...');
         var d = connectionData;
-        connect(d[0], d[1], d[2], d[3], d[4], function() {
+        // won't work if totp is mandetory
```
Techman: Slight spelling error here, correct spelling is 'mandatory'.

AStove: ty

index.html
```diff
@@ -123,18 +123,26 @@ <h4 class="panel-title">
...
+                        <input type="checkbox" id="autoconnect" ng-model="settings.autoconnect"  ng-disabled="settings.useTotp">
                         Automatically connect
                     </label>
                   </div>
                   <div class="checkbox">
-                    <label class="control-label" for="ssl">
-                      <input type="checkbox" id="ssl" ng-model="settings.ssl">
-                      Encryption. <strong>Strongly recommended!</strong> Need help? Check below.
+                    <label class="control-label" for="useTotp">
+                      <input type="checkbox" id="useTotp" ng-model="settings.useTotp" ng-change="changeUseTOTP()">
```
lorenzhs: you know you can add a callback for when a setting changes, too? https://github.com/glowing-bear/glowing-bear/blob/3d042b125bd40f51f4d9cc53fa9abb801f11cd71/js/settings.js#L52-L65

lorenzhs: Using that is a bit cheaper on the CPU and more robust because it also triggers if any other code changes the setting, not just when the checkbox is changed

index.html
```diff
@@ -123,18 +123,26 @@ <h4 class="panel-title">
...
+                      Use Time-based One-Time Password
```
lorenzhs: This probably needs an :information_source: link next to it to the weechat docs or some kind of explanation

lorenzhs:
Thanks, you've implemented a lot more of the things that were missing than I did :)

I kinda prefer the design of #1040 though, where the token input is next to the password input like host and port. This makes the dynamic switching a bit more complicated (need to add/remove classes), but what do you think?

Also, I don't have a test instance at hand right now so I can't check, but how do you handle reconnects? I guess they should simply be disabled?

index.html
```diff
@@ -123,18 +123,26 @@ <h4 class="panel-title">
...
                     </label>
                   </div>
+                  <label class="control-label" for="totp" ng-Show="settings.useTotp">One time password</label>
+                  <input type="text" class="form-control favorite-font" id="totp" ng-model="totp" ng-Show="settings.useTotp" ng-model="totp" autocomplete="off">
```
lorenzhs: you've specified `ng-model` twice here

index.html
AStove: I did not know, thanks. Changed.

AStove:
-Changed the layout to match #1040, also separated the port and hostname label this seemed fitting.
-Added information link next to totp checkbox
-Don't attempt reconnect when useTotp setting is used
-Validate totp

Loosely related: I noticed weechat accepts totp argument even when totp is not enabled. I made a [feature request](https://github.com/weechat/weechat/issues/1435) to this for weechat.

index.html
```diff
@@ -123,20 +133,26 @@ <h4 class="panel-title">
...
                         Automatically connect
                     </label>
                   </div>
                   <div class="checkbox">
-                    <label class="control-label" for="ssl">
-                      <input type="checkbox" id="ssl" ng-model="settings.ssl">
-                      Encryption. <strong>Strongly recommended!</strong> Need help? Check below.
+                    <label class="control-label" for="useTotp">
+                      <input type="checkbox" id="useTotp" ng-model="settings.useTotp">
+                      Use Time-based One-Time Password <a href="https://blog.weechat.org/post/2019/01/14/Support-of-TOTP">ℹ️</a>
```
lorenzhs: Link should be `target="_blank"` I guess, but this applies to the other links as well, so we should just change it everywhere at some point

lorenzhs: Also maybe more prominent. Bootstrap comes with some icons, maybe use those:
```html
<i class="glyphicon glyphicon-info-sign"></i>
```

index.html
```diff
@@ -123,20 +133,26 @@ <h4 class="panel-title">
                         Save password in your browser
                     </label>
                   </div>
+                  <div class="checkbox">
+                    <label class="control-label" for="ssl">
+                      <input type="checkbox" id="ssl" ng-model="settings.ssl">
+                      Encryption. <strong>Strongly recommended!</strong> Need help? Check below.
+                    </label>
+                  </div>
```
lorenzhs: Why reorder the inputs like that? It's weird to have save password -> encryption -> autoconnect when autoconnect depends on save password

js/glowingbear.js
```diff
@@ -687,6 +688,16 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
...
     };
 
+    settings.addCallback('useTotp', function() {
+        if (settings.useTotp) {
+            settings.autoconnect = false;
+        }
+    });
+
+    $scope.parseTotp = function() {
+        $scope.totpInvalid = !/^\d{6}$/.test($scope.totp);
```
lorenzhs: From the linked weechat blog post:

> the number of digits (4 to 10 digits are supported, 6 is the default and recommended value).

lorenzhs:
Thanks, I really like this :)

index.html
AStove: Agree but I didn't specifically reorder. I initially added it to the bottom and then moved the input to the top.. Better?

lorenzhs:
Thanks a lot!

Issue #415, opened 2014-08-20
Allow for resizing of buffer and nick lists


lorenzhs:
- store their sizes in a setting
- modify CSS in JS
- dragging them to size would be nicest

PR #1099, opened 2019-12-22
Flex layout with boostrap4

### LINKED ISSUE #873: top(ic) overlays scroll bars
The top bar with the glowing-bear logo, the channel topic as well as the settings and log out button, overlays both the top end of the scroll bars: The one for the list of buffers, the current buffer and the one for the nick list. Of course provided the mentioned element/list is long enough to have a scroll bar displayed.

It's mostly a surficial inconvenience, but with a very long list, it is possible to completely lose the scroll bar under the top bar, which makes grabbing it with the mouse impossible.

### LINKED ISSUE #888: Long links stretch mobile view
When long hyperlink appears in buffer it stretches window a bit (part of the text is still visually cut), wider element makes all lines in buffer wrap a bit further (off screen) and causes horizontal scrolling to be possible.
I'm not sure how long words should be handled (as wrapping lines without spaces in them isn't an easy job) but it's not only annoying but also seems to slow down app.
Happens on latest edge, can't test it on droid or iPhone

### LINKED ISSUE #1066: Can't tile to quarter-screen on 1080p monitor
I use quarter-tiling a lot, where many messaging apps take up a quarter of the screen.

With Glowing Bear (built in Electron) once I get close to quarter size it starts to cut off content on the far right-hand side. Any chance we could just keep wrapping content?

AStove:
fixes #1066 
fixes #888
fixes #873

Using bootstrap and their flex layout for the main interface. This PR shouldn't do any bugfixes or features for now. (And frankly it will create some bugs.) But this will enable us to improve things in the future.

Anyone noticing things, please comment and I'll keep the todolist up to date.

Reason:
I was looking to implement a small feature: making the side menus dragable (#415). It was then that I noticed how the interface was made was far from ideal. A lot of media queries were used and everything with fixed sized and `position: fixed`. Making it very hard to resize one element and expecting other elements to fill int he space, etc.

Upsides:
- Flexibility of the interface towards sizing
- Reduce codebase slightly
- More modern libraries

Downsides:
- Possible introduction of bugs
- Size increase due to jQuery and Bootstrap
  - Only for first loads, otherwise it's chached
- Performance of flex layout?
  - Seems similar
- IE11 no longer works

Major changes:
These were things that work differently in bootstrap 4 and couldn't be avoided
- JQuery slim used
- Font Awesome used (glyphicons no longer included in bootstrap)
- Panels become cards (for the accordion)
- Modals slight differences

Todo:
- ~~Slide in-out animations for side menus~~
- ~~Check light and base16 theme~~
- ~~General testing~~
- ~~Test image drag and drop~~
- ~~Testing on mobile~~
- Testing all browsers
  - ~~Chrome~~
  - ~~Firefox~~
  - Safari
  - IE11 doesn't work but does anyone care? Before these changed it worked.
  - Edge
- ~~Further prune css~~
- ~~Check plugins, the basic ones seem to work at first glance~~
- ~~Make nicklist hide on too small screens~~
- ~~Check alerts~~
- ~~'pull right' isnt supported anymore so eradicate that, plugins still use it.~~
- ~~Font of nicklist and buffer list~~
- ~~Size of the password field~~
- ~~The "important note" is not visible for non-javascript users~~
- ~~Channel title isn't centered on small screens~~

### DIFF
--- a/index.html
+++ b/index.html
@@ -238,15 +238,15 @@
               <p>Nearly done! Now you just need to copy the files into place. To do that, use the following commands,
                 replacing the <strong>username</strong> placeholder with your actual username:</p>
                 <pre><code>mkdir -p ~<strong>username</strong>/.weechat/ssl
-  cat /etc/letsencrypt/live/{{ settings.host || your.domain.com }}/{fullchain,privkey}.pem &gt; ~<strong>username</strong>/.weechat/ssl/relay.pem
-  chown -R <strong>username</strong>:<strong>username</strong> ~<strong>username</strong>/.weechat/ssl/</code></pre>
+cat /etc/letsencrypt/live/{{ settings.host || your.domain.com }}/{fullchain,privkey}.pem &gt; ~<strong>username</strong>/.weechat/ssl/relay.pem
+chown -R <strong>username</strong>:<strong>username</strong> ~<strong>username</strong>/.weechat/ssl/</code></pre>
               <p>Once you've got the certificate and moved it in place, you can set up an encrypted relay on port
                 {{ settings.port || 9001 }} with these WeeChat commands:</p>
               <pre>
-  /set relay.network.password y0ur_StRonG-pa$sw0rd:of*choice
-  /relay sslcertkey
-  /relay add ssl.weechat {{ settings.port || 9001 }}
-  </pre>
+/set relay.network.password y0ur_StRonG-pa$sw0rd:of*choice
+/relay sslcertkey
+/relay add ssl.weechat {{ settings.port || 9001 }}
+</pre>
               <p>Your certificate needs to be renewed every couple of months. Either follow the instructions for automatic
                 renewal at <a href="https://certbot.eff.org/">https://certbot.eff.org</a>, or run
                 <code>certbot renew</code> manually when renewal is due. <strong>Important:</strong> You'll need to follow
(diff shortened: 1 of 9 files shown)


lorenzhs:
One issue I observed is that the global unread badge isn't properly displayed:
Screenshot from 2019-12-31 18-01-58

Other than that, a lot of sizes and spacing has changed (eg, the height of bufferlist and nicklist items, the width of the nicklist, ...), is that intentional?

index.html
```diff
@@ -1,633 +1,715 @@
 <!DOCTYPE html>
 <html ng-app="weechat">
-  <head>
```
lorenzhs: why did you reindent everything? makes it very hard to see what's changed

AStove:
It's not intentional bootstrap 4 has other default margins and such. I tried to adjust some things but it took longer than I wanted it to take and decided to do a pull request with what I had and hope for more contributions. I'm contemplating closing this pull request. The desired goals don't warrant the risks of this big change.

index.html
AStove: At one point I was stuck with a div and I hit a magic key combination in VS Code to fix all indentation, I shouldn't have done that. However a lot did change legitimatly too because it ended up in a new parent div. I changed the indentation of the head and body back to what it was now, this make is slightly better.

AStove:
Badges fixed

AStove:
Too many downsides vs upsides at the moment.

PR #1102, opened 2020-01-25
Added support for Imgur token (authed upload) and Imgur albums.

Jaekr:
Added:

- Authed upload: add a Imgur Bearer Token to upload directly into your Imgur account
- Album upload (need to be authenticated): uploads the image and adds it into the selected album

### DIFF
--- a/index.html
+++ b/index.html
@@ -123,7 +123,7 @@
                   <div class="input-group">
                     <div class="row no-gutter">
                       <div class="col-sm-9">
-                        <label class="control-label" for="iAlb">Imgur Album hash<a target="_blank" href="https://github.com/Jaekr/glowing-bear/wiki/Getting-a-token-&-Album-Hash"><i class="glyphicon glyphicon-info-sign"></i></a></label>
+                        <label class="control-label" for="iAlb">Imgur Album hash<a target="_blank" href="https://github.com/glowing-bear/glowing-bear/wiki/Getting-an-imgur-token-&-album-hash"><i class="glyphicon glyphicon-info-sign"></i></a></label>
                         <input type="text" class="form-control favorite-font" id="iAlb" ng-model="settings.iAlb" placeholder="Album hash" autocapitalize="off">
                       </div>
                     </div>
--- a/index.html
+++ b/index.html
@@ -100,7 +100,6 @@
             <div class="panel-body">
                <form class="form-signin" role="form">
                 <div class="form-group">
-
                   <div class="input-group">
                     <div class="row no-gutter">
                       <div class="col-sm-9">
@@ -110,6 +109,22 @@
                       <div class="col-sm-3">
                         <label class="control-label" for="port">Port</label>
                         <input type="text" class="form-control favorite-font" id="port" ng-model="settings.port" ng-disabled="portDisabled" placeholder="Port">
+                      </div>
+                    </div>
+                  </div>
+                  <div class="input-group">
+                    <div class="row no-gutter">
+                      <div class="col-sm-9">
+                        <label class="control-label" for="iToken">Imgur Access token</label>
+                        <input type="text" class="form-control favorite-font" id="iToken" ng-model="settings.iToken" placeholder="Token" autocapitalize="off">
+                      </div>
+                    </div>
+                  </div>
+                  <div class="input-group">
+                    <div class="row no-gutter">
+                      <div class="col-sm-9">
+                        <label class="control-label" for="iAlb">Imgur Album hash<a target="_blank" href="https://github.com/Jaekr/glowing-bear/wiki/Getting-a-token-&-Album-Hash"><i class="glyphicon glyphicon-info-sign"></i></a></label>
+                        <input type="text" class="form-control favorite-font" id="iAlb" ng-model="settings.iAlb" placeholder="Album hash" autocapitalize="off">
                       </div>
                     </div>
                   </div>
--- a/index.html
+++ b/index.html
@@ -112,22 +112,6 @@
                       </div>
                     </div>
                   </div>
-                  <div class="input-group">
-                    <div class="row no-gutter">
-                      <div class="col-sm-9">
-                        <label class="control-label" for="iToken">Imgur Access token</label>
-                        <input type="text" class="form-control favorite-font" id="iToken" ng-model="settings.iToken" placeholder="Token" autocapitalize="off">
-                      </div>
-                    </div>
-                  </div>
-                  <div class="input-group">
-                    <div class="row no-gutter">
-                      <div class="col-sm-9">
-                        <label class="control-label" for="iAlb">Imgur Album hash<a target="_blank" href="https://github.com/glowing-bear/glowing-bear/wiki/Getting-an-imgur-token-&-album-hash"><i class="glyphicon glyphicon-info-sign"></i></a></label>
-                        <input type="text" class="form-control favorite-font" id="iAlb" ng-model="settings.iAlb" placeholder="Album hash" autocapitalize="off">
-                      </div>
-                    </div>
-                  </div>
                   <div class="row no-gutter">
                     <div ng-class="settings.useTotp ? 'col-sm-9' : 'col-sm-12'" ng>
                       <label class="control-label" for="password">WeeChat relay password</label>
@@ -473,7 +457,7 @@ npm run build-electron-{windows, darwin, linux}</pre>
                     </div>
 
                     <label for="size" class="col-sm-1 control-label">Size</label>
-                    <div class="col-sm-2">
+                    <div class="col-sm-3">
                       <input type="text" ng-model="settings.fontsize" class="form-control" id="size">
                     </div>
                   </div>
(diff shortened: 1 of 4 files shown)


js/imgur.js
```diff
@@ -26,8 +26,40 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+        // To get the albumHash, go to imgur.com and create a new album / go to an existing one and pick the string in the url: https://imgur.com/a/<string>
+        //
+
+        var accessToken = "";
+        var albEnabled = "";
+        var albumHash = "";
+        var isClientID = "";
+
+        // Set client ID (Yours)
+        if(localStorage.getItem("iToken").length < 38){
```
lorenzhs: using localStorage directly here isn't wrong per se, but we have a settings module (js/settings.js) which we use for all kinds of settings stuff. I think you should use that one instead to keep things uniform.

js/imgur.js
```diff
@@ -26,8 +26,40 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+        // Getting the access token
+        //
+        // First, you need a client id, go here to register an app: https://api.imgur.com/oauth2/addclient
+        //
+        // Go to: https://api.imgur.com/oauth2/authorize?client_id=<client id>&response_type=token&state=askin-01
+        //
+        // Accept and it sould redirect you with an url like this: https://<website.tld>/?state=askin-01#access_token=<token here>&expires_in=315360000&token_type=bearer&refresh_token=<refresh token>&account_username=<your nick>&account_id=<your id>
+        //
+        // You can now input this token in the main page to get images uploaded to your account
+        // To get the albumHash, go to imgur.com and create a new album / go to an existing one and pick the string in the url: https://imgur.com/a/<string>
```
lorenzhs: This should probably go in the settings dialog as a help text for the input fields, as a hover text for a ":question:" badge next to the label or something like that. Users aren't going to look in the code.

js/imgur.js
```diff
@@ -61,6 +102,7 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
             // Remove progress bar
             currentProgressBar.parentNode.removeChild(currentProgressBar);
 
+            console.log(xhttp.status);
```
lorenzhs: this should probably be removed before we merge

js/imgur.js
```diff
@@ -26,8 +26,40 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+        // Accept and it sould redirect you with an url like this: https://<website.tld>/?state=askin-01#access_token=<token here>&expires_in=315360000&token_type=bearer&refresh_token=<refresh token>&account_username=<your nick>&account_id=<your id>
+        //
+        // You can now input this token in the main page to get images uploaded to your account
+        // To get the albumHash, go to imgur.com and create a new album / go to an existing one and pick the string in the url: https://imgur.com/a/<string>
+        //
+
+        var accessToken = "";
+        var albEnabled = "";
+        var albumHash = "";
+        var isClientID = "";
```
lorenzhs: Why don't you set the token to our client ID by default, `isClientID = true`, and only set it to the user's access token and `false` if the settings are there? Would be much easier.

js/imgur.js
```diff
@@ -26,8 +26,40 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+
+        // Set client ID (Yours)
+        if(localStorage.getItem("iToken").length < 38){
+            accessToken = "164efef8979cd4b"; // Glowing bear default token
+            isClientID = 'true';
```
lorenzhs: In any case, you should be using booleans (`true` and `false`) instead of strings (`'true'` and `'false'`) for `isClientID`

js/imgur.js
```diff
@@ -45,14 +77,23 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
         fd.append("image", base64img); // Append the file
         fd.append("type", "base64"); // Set image type to base64
 
+        // Comment the next line if you don't want to upload your pics into a folder.
+        if(albEnabled == 'true') {
```
lorenzhs: If you use booleans as mentioned above, this would just be `if (albEnabled) { /* ... */ }`

js/imgur.js
```diff
@@ -45,14 +77,23 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+
         // Create new XMLHttpRequest
         var xhttp = new XMLHttpRequest();
 
         // Post request to imgur api
         xhttp.open("POST", "https://api.imgur.com/3/image", true);
 
         // Set headers
-        xhttp.setRequestHeader("Authorization", "Client-ID " + clientId);
+        if(isClientID == 'false') {
```
lorenzhs: Same here,  `if (isClientID) { /* ... */ } else { /* ... */ }`

js/imgur.js
```diff
@@ -45,14 +77,23 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
         fd.append("image", base64img); // Append the file
         fd.append("type", "base64"); // Set image type to base64
 
+        // Comment the next line if you don't want to upload your pics into a folder.
```
lorenzhs: This comment doesn't make sense:
- users don't change the code
- if you don't want to upload your pics into a folder, just don't set the album hash.

lorenzhs:
Thanks @Jaekr, this is a nice addition. I already added a few inline comments, but there's one other thing. I would suggest moving the input fields into the settings dialog instead of the connection screen. For one, it doesn't have anything to do with connecting GB, and for another, if it's in settings, you can change it without reconnecting.

Please also don't commit the package-lock.json, npm is only used for the test environment and we don't really care about exact versions of anything.

js/imgur.js
```diff
@@ -26,8 +26,40 @@ weechat.factory('imgur', ['$rootScope', function($rootScope) {
...
+        var albEnabled = "";
+        var albumHash = "";
+        var isClientID = "";
+
+        // Set client ID (Yours)
+        if(localStorage.getItem("iToken").length < 38){
+            accessToken = "164efef8979cd4b"; // Glowing bear default token
+            isClientID = 'true';
+        } else {
+            accessToken = localStorage.getItem("iToken").replace(/['"]+/g, '');;
```
lorenzhs: There's an extra semicolon here at the end of the line: `;;`

Jaekr:
I'm gonna make the changes as fast as possible!

js/imgur.js
Jaekr: Corrected

js/imgur.js
Jaekr: Corrected!

js/imgur.js
Jaekr: Now corrected.

js/imgur.js
Jaekr: Corrected.

js/imgur.js
Jaekr: Corrected.

js/imgur.js
Jaekr: Corrected.

index.html
```diff
@@ -113,6 +113,29 @@ <h4 class="panel-title">
                       </div>
                     </div>
                   </div>
+
+                  <!-- Start add -->
```
lorenzhs: You don't need these comments, they make no sense once merged

Jaekr: Corrected.

js/imgur.js
Jaekr: Should work.

js/imgur.js
Jaekr: Should be good to go!

Jaekr:
That's all, it should be ok.

lorenzhs:
All right, thanks a lot! I squashed it a bit and moved the configuration to the settings dialog, you can test the result at https://latest.glowing-bear.org/pull/1102/

js/imgur.js
```diff
@@ -3,7 +3,7 @@
 
 var weechat = angular.module('weechat');
 
-weechat.factory('imgur', ['$rootScope', function($rootScope) {
+weechat.factory('imgur', 'settings', ['$rootScope', function($rootScope) {
```
AStove: Because of this I'm getting `angular.js:15536 Error: [ng:areq] http://errors.angularjs.org/1.7.5/ng/areq?p0=fn&p1=not%20a%20function%2C%20got%20string`

See https://stackoverflow.com/questions/21673404/error-ngareq-from-angular-controller

You should not add the 'settings' there.

Jaekr: I'm testing that.

AStove: Do 

`weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, settings) {`

Jaekr: Now fixed.

Jaekr:
Any last changes?

lorenzhs:
Thanks very much @Jaekr, this is a feature that's been requested a few times, I think you just made a few people quite happy :)

PR #1104, opened 2020-02-19
Show a toast when clicking nick of a user who left

lorenzhs:
Adds a self-dismissing popover when clicking the nick of a user who left in the chat history

screenshot

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -715,7 +715,6 @@ img.emojione {
   width: 250px;
   margin-left: -125px;
 
-  background-color: #333;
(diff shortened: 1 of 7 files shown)


AStove:
I tested it a bit, the only thing I could find was that it's unreadable in light theme.

lorenzhs:
good catch, thanks! needs a different background in the light theme

lorenzhs:
I added some styling for all the other themes (light, blue, and the various solarized ones). Blue even got a border and stuff to make it fit with the rest of the theme.

PR #1109, opened 2020-04-05
Add TikTok plugin

AStove:
Embeds tik tok links.. Is very similar to twitter. The embed is rather big but I don't see a simple way of changing it.

https://developers.tiktok.com/doc/Embed

example link: https://www.tiktok.com/@scout2015/video/6718335390845095173

Only those links are embedded. Tiktok also has shortened links in the format vm.tktok.com/xxxx. Those are impossible to embed in my experience due to CORS. They are a redirect to a normal link but it's not possible to get the redirect location in code due to CORS.

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -558,12 +558,14 @@ plugins.factory('userPlugins', function() {
      * Very similar to twitter
      */
     var tikTokPlugin = new UrlPlugin('TikTok', function(url) {
-        var regex = /^https?:\/\/(www\.)?tiktok\.com\/(.+)/i,
+        var regex = /^https?:\/\/(www\.)?tiktok\.com\/@(.+)\/video\/(.+)/i,
             match = url.match(regex);
 
         if (match) {
 
             return function() {
+                var element = this.getElement();
+                
                 fetch("https://www.tiktok.com/oembed?url=" + url)
                 .then(function(response) {
                     return response.json();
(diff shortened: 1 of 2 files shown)


lorenzhs:
Why is your initial regex so broad? It probably shouldn't match links like `https://www.tiktok.com/legal/privacy-policy?lang=en`

js/plugins.js
```diff
@@ -553,8 +553,47 @@ plugins.factory('userPlugins', function() {
...
+
+        if (match) {
+
+            return function() {
+
+                var element = this.getElement();
+
+                GetEmbedContent( "https://www.tiktok.com/oembed?url=" + url, element);
+
+                function GetEmbedContent(embedurl, element) {
```
lorenzhs, line 572: Is there a reason that you turned this into a separate function that's used in exactly one place?

js/plugins.js
```diff
@@ -553,8 +553,47 @@ plugins.factory('userPlugins', function() {
...
+                    fetch(embedurl)
```
lorenzhs, line 573: Apparently IE11 doesn't support fetch. I guess for the three people still using that the plugin just wouldn't work? Since the rest of GB would still work for them, I guess that's fine.

js/plugins.js
AStove: Fixed, it had a reason but it did not play out because I could not get the vm.tiktok.com links to work.

AStove:
> Why is your initial regex so broad? It probably shouldn't match links like `https://www.tiktok.com/legal/privacy-policy?lang=en`

Same goes for someother other plugins, some examples:
https://soundcloud.com/charts/top
https://www.mixcloud.com/terms/
I'll make an issue of it.

Fixed, it's more specific now.

AStove:
> Edit: oh, some tests would be great, too

Done.

lorenzhs:
Many thanks!

PR #1112, opened 2020-04-09
allow to set timezone in settings

### LINKED ISSUE #1111: No way to set timezone for timestamp
I work for a global company, we use UTC for work, my weechat server is also in UTC.
However, in glowing-bear, it's always showing time in local/browser timezone.
We need a setting to specify timezone, at least allow me to switch between local time and UTC.

guoqiao:
currently glowing-bear always show message timestamp in browser timezone.
It's common that people want to use different tz like UTC in irc.

In current code, we are using `date` filter in angular for time.
It has a 3rd `timezone` param but ignored, which will use brower timezone:

    https://docs.angularjs.org/api/ng/filter/date

This patch add a new `timezone` option in settings, and pass it to date filter.
It doesn't change/break anything when blank, but we have he ability to change timezone now.

Closes #1111


AStove:
I filled in "+0430", reloaded the page. Get javascript errors. When I look in the setting again it now says "430" so I'm guessing it converted to int.

`TypeError: a.replace is not a function
    at dc (angular.js:1460)
    at angular.js:22978
    at new BufferLine (models.js:380)
    at handleLine (handlers.js:140)
    at handlers.js:370
    at Array.forEach (<anonymous>)
    at Object.handleLineInfo (handlers.js:369)
    at connection.js:513
    at angular.js:17914
    at m.$digest (angular.js:19075) "Possibly unhandled rejection: {}"`

guoqiao:
@AStove  actually when I was debuging, I had the same issue. But after I tried different input and reload glowing-bear page a few times, the issue is gone.  Some clue I noticed:
1. string like "UTC", "CST" will always work.
2. for numbers, If you try "+0400" instead of "+0430", it does work, but not all the time. I feel like angular is doning something weird here.

AStove:
It doesn't work for "+0400" either for me. The stored value in localstorage is correctly a string "+0400". But it still gets converted to "400". [Seems like there have been problems with this before.](https://github.com/glowing-bear/glowing-bear/blob/c7a5a65ed74f4a084666e0d32b8aac0fe4002fd6/js/settings.js#L9)

AStove:
The conversion mistake happens in [settings.js](https://github.com/glowing-bear/glowing-bear/blob/c7a5a65ed74f4a084666e0d32b8aac0fe4002fd6/js/settings.js#L25)

guoqiao:
@AStove  I think I find out how to reproduce it:
If you change the timezone setting and save, DO NOT reload glowing-bear, instead, reload a buffer via switching to another one and switch back, the timezone setting is working fine. And if you open settings modal or check window.localStorage, you can see timezone is correct with "+0430".
However, if you reload glowing-bear, the issue will happen.
I think the problem is form this line: https://github.com/glowing-bear/glowing-bear/blob/c7a5a65ed74f4a084666e0d32b8aac0fe4002fd6/js/settings.js#L25

when settings module load value from $store, the implicit type cast happened there.

AStove:
It works if you use "'+0400'" so with the quotes typed into the box. Store it with quotes?

guoqiao:
@AStove  that's a good work around.  It will be even better if we can find out why the $store.get func did the implicit type cast and how we can avoid it.

guoqiao:
@AStove  I find out the issue, which is from this function:
https://github.com/glowing-bear/glowing-bear/blob/c7a5a65ed74f4a084666e0d32b8aac0fe4002fd6/js/localstorage.js#L68

`JSON.parse(res)` and `if (parseFloat(val) == val ...`.

guoqiao:
`parseFloat(val) == val` is the one to blame, or to be precise, `==` in JS:
```
var x = "+0430";
var y = parseFloat(x)  --> 430
x == y --> true
```

guoqiao:
So TIL:
In Javascript, `a == b` will first try to convert a and b to same type, and then compare.
To do a real compare, use  `a === b`.
I will raise another bug for this bit.

AStove:
Change that in a different pull request, this might break a lot and has essentially nothing to do with this timezone zetting.

lorenzhs:
That's the point in this case, though. If we did `if (parseFloat(val).toString() === val)` then `12.0` wouldn't be recognised as a number any more.

guoqiao:
> That's the point in this case, though. If we did `if (parseFloat(val).toString() === val)` then `12.0` wouldn't be recognised as a number any more.

I think the localstorage module should not do this tricky thing at all(gussing type, in wrong way, even use `===`), it should just store anything as it is.

For example, when I store `123456`,  you never know it's a number or it's my password.
Only the code who uses this value knows it,  and should do the type cast there.

lorenzhs:
I don't think adding a setting for this is a good solution. It's such a niche use case that adding a setting clearly isn't the way to go about this. We don't have settings for a million other things that someone's use case required. In most cases, we found a better solution in the end that doesn't need a setting. We're currently figuring out if/how we can get weechat to tell us its timezone.

AStove:
You could use this extension https://chrome.google.com/webstore/detail/spoof-timezone/kcabmhnajflfolhelachlflngdbfhboe?hl=en

And limit it only to the glowing bear url.
image

I just tested it and it works well at first glance. Additionally if you have other domains you wish to see in UTC time, you can.

lorenzhs:
@AStove that's an excellent suggestion!

guoqiao:
@AStove  ok, good enough, thanks.

PR #1114, opened 2020-04-12
Hide bufferline times for relay buffers

AStove:
For certain buffers a timestamp does not make sense and is always 00:00:00. This is the case for the buffertype 'relay'. This hides the timestamps for these buffers.

TElyS4YsF1

### DIFF
--- a/index.html
+++ b/index.html
@@ -407,12 +407,12 @@ npm run build-electron-{windows, darwin, linux}</pre>
           </tbody>
           <tbody ng-repeat="bufferline in bufferlines">
             <tr class="bufferline">
-              <td class="time" ng-show="activeBuffer().showTime">
+              <td class="time" ng-show="activeBuffer().showBufferLineTimes">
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -407,12 +407,12 @@ <h4 class="panel-title">
           </tbody>
           <tbody ng-repeat="bufferline in bufferlines">
             <tr class="bufferline">
-              <td class="time">
+              <td class="time" ng-show="activeBuffer().showBufferLineTimes">
```
lorenzhs, line 410: This should use one-time bindings (`::` prefix) because it never changes once set. It should also be a class of the `bufferlines` div like `withNicklist` that affects all `td.time` and `td.prefix` beneath it, because it affects all lines of a buffer.

AStove: I did it with a class, I didn't get the one time binding in there because each time the buffer chages this should be updated.

lorenzhs:
Thanks!

index.html
lorenzhs: Yeah once it's on the table object it needs to be a watcher, but it's only a single one so no problem

PR #1118, opened 2020-04-13
Show correct error message when giving wrong password in firefox

### LINKED ISSUE #1116: Wrong error message when connecting with wrong password in firefox
Issue reported by  @yourfate

Connecting with the wrong password gives the message "Connection error" and "Secure connection error" instead of "Error: wrong password or token".

Conditions:
- Use encryption
- Use firefox
- Use the wrong password

Chrome (correct):
image

Firefox (wrong):
image

AStove:
fixes #1116 
Change the way we initialize, after sending the init message, give it 100ms for weechat to check the password and see if weechat closed the connection or not. If after 100ms the connection is still open we proceed to send the info version command.

What the problem is with sending info version instantly is that chrome handles it differently than firefox. Chrome has already sent the info version before the connection is closed. While firefox send it right after the connection is closed and generates an error. This error is assumed to be a certificate error so the wrong error message is shown to the user.

The indents might not be correct around connection.js ln 206 but that is for clarity of the changes.

Checks I make:
- firefox gives right error when password is wrong
- chrome gives right error when password is wrong
- firefox gives right error when certificate is wrong
- chrome gives right error when certificate is wrong

### DIFF
--- a/js/connection.js
+++ b/js/connection.js
@@ -38,6 +38,7 @@ weechat.factory('connection',
             var _initializeConnection = function(passwd) {
                 // Escape comma in password (#937)
                 passwd = passwd.replace(',', '\\,');
+
                 // This is not the proper way to do this.
                 // WeeChat does not send a confirmation for the init.
                 // Until it does, We need to "assume" that formatInit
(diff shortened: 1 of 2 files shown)


js/connection.js
```diff
@@ -51,6 +51,22 @@ weechat.factory('connection',
...
+                // Give weechat the oportunity to close the connection before we send anything else
+                return new Promise(function(resolve, reject) {
+                    setTimeout(function() {
+                        //If it's still open after 100ms we'll assume we are connected
+                        if (ngWebsockets.readyState() === WebSocket.OPEN) {
+                            resolve();
+                        } else {
+                            reject();
+                        }
+                    },100);
```
lorenzhs, line 65: I think it should be documented that this is a bit of a magic number, but that it's there only to improve error communication

lorenzhs:
This doesn't fix the issue when "Automatically connect" is enabled, I still get the old (wrong) message. Any ideas why?

AStove:
Recent developments in weechat are that a handshake will be provided. This problem should be fixed in the implementation of this new handshake. Therefor I will close this PR and mention the issue in a next PR to be expected.

PR #1119, opened 2020-04-17
Use new authentication methods in weechat 2.9

### LINKED ISSUE #1116: Wrong error message when connecting with wrong password in firefox

AStove:
https://weechat.org/files/doc/devel/weechat_relay_protocol.en.html
To test this you will need the latest weechat 2.9-dev version off github and build it yourself for now.

- Add new handshake and init functions of weechat 2.9
- Add checkbox to allow plaintexts passwords, this is checked by default to allow older versions. Perhaps later when the version has been out for a while this can be turned off by default.
- Move the checkboxes around for more logical UI.
- Remove the TOTP field and use a prompt instead.
- Only support pbkdf2+sha512, this is the most secure, no need to implement multiple imo
- Old version error to say an old weechat version was detected and to check the "allow plaintext authentication" checkbox.
- Hash algorithem disagree if the hashing algorithem wasn't set properly in weechat

Testing done
========
Windows Desktop Chrome:
- Still works with old weechat
- Still works with old weechat with TOTP
- Works with new version of weechat
- Works with new version of weechat with TOTP

What we can still add is to remember that a server is old, and then skip the handshake altogether. This would save 1second connection time. However I don't think we should keep doing the plaintext password if weechat gets updated and the user does not uncheck the "allow plaintext authentication" checkbox.

Also on bad connections, there should be an option to wait longer for the handshake response, perhaps put a button in the "old version" error to try for longer?

fixes #1116


AStove:
Updated to the latest weechat version where the names in the handshake were changed a little.

index.html
```diff
@@ -68,6 +68,12 @@ <h2>
       <div class="alert alert-danger" ng-show="securityError" ng-cloak>
         <strong>Secure connection error</strong> Unable to connect to unencrypted relay when you are connecting to Glowing Bear over HTTPS. Please use an encrypted relay or load the page without using HTTPS.
       </div>
+      <div class="alert alert-danger" ng-show="oldWeechatError" ng-cloak>
+        <strong>Weechat version error</strong> Weechat connected but did not respond to a handshake. This could mean weechat < version 2.9. Verify your weechat version and check "Allow Plaintext Authentication" if it's < 2.9.
+      </div>
+      <div class="alert alert-danger" ng-show="hashAlgorithemDisagree" ng-cloak>
+        <strong>Hash algorithem error</strong> Weechat and glowing bear did not agree on a hashing algorithem, please do /set relay.network.password_hash_algo "pbkdf2+sha512" in weechat.
```
lorenzhs, line 75: The word is "algorithm" without an "e" ;)

js/connection.js
```diff
@@ -31,31 +34,143 @@ weechat.factory('connection',
...
+                        key, //your key from generateKey or importKey
+                        512
+                    );
+                    
+                }).then( function (hash) {
+
+
+                    ngWebsockets.send(
+                        weeChat.Protocol.formatInit29(
+                            'pbkdf2+sha512:' + utils.bytetoHexString(salt) + ':100000:' + utils.bytetoHexString(hash),
```
lorenzhs, line 160: The iterations probably shouldn't be hardcoded here

js/weechat.js
```diff
@@ -628,17 +628,51 @@
...
+        var defaultParams = {
+            password_hash_algo: 'pbkdf2+sha512',
+            compression: 'zlib'
+        };
+        var keys = [];
+        var parts = [];
+
+        params = WeeChatProtocol._mergeParams(defaultParams, params);
+
+        if (params.password !== null) {
```
lorenzhs, line 649: why do these depend on the password not being null?

lorenzhs:
Thanks, this looks pretty cool! I haven't tested it yet, and clearly there's a lot of stuff that needs to be tested. But I think the option shouldn't be called "Allow Plaintext Authentication (Weechat < 2.9)", that exposes the internals too much and "plain text" isn't especially accurate because transport encryption is orthogonal to this setting, and sending the password via an encrypted channel isn't plain text. I think it should be called something like "Compatibility with Weechat 2.8 and older".

js/connection.js
AStove: Yes, sorry, fixed now. Tested to see if it still works and it does.

js/weechat.js
AStove: Two seperate mistakes, "password" used to be what "password_hash_algo" is now, forgot to change that. And the compression should always have been compression.

Updated, fixed, tested.

index.html
AStove: Agree to dissagree, That's just too many consonants in a row... ;) (fixed all occurenced)

AStove:
I renamed the new setting. Thanks for the reviews so far.

lorenzhs:
How far along would you say this PR is? Is it mostly done and just needs some testing/reviewing, or is there more left to do?

AStove:
IMO it's mostly done, I can't think of much more features to implement that would be of use. Perhaps some testing on other browsers/electron is needed. I only tested it on windows chrome, firefox and android chrome.

Either way no need to merge before weechat 2.9 is out?

AStove:
I forgot, what do you think about that feature to "remember it's an old server" and to set the handshake timeout shorter or skip the handshake altogether.

lorenzhs:
I think that would be a good idea, but it can be a separate pull request. But at this point (i.e., until a few months after 2.9 is released), I think waiting 2s is way too long. It's a major inconvenience. It should probably not be more than 200ms initially (would have to test out how it feels, 200ms is just a guess). We should increase it once 2.9 becomes more widespread, but having it too high to soon would just irritate a lot of users for no real benefit.

index.html
```diff
@@ -113,39 +119,41 @@ <h4 class="panel-title">
                     </div>
                   </div>
                   <div class="row no-gutter">
-                    <div ng-class="settings.useTotp ? 'col-sm-9' : 'col-sm-12'" ng>
+                    <div ng-class="col-sm-12" ng>
```
lorenzhs, line 122: should just be `class`, not `ng-class`

AStove:
https://github.com/weechat/weechat/releases

Weechat Pre-Release of 2.9 is out, we can now test this.

lorenzhs:
@torhve are we good to merge this?

PR #1122, opened 2020-04-28
Command completion for weechat 2.9

### LINKED ISSUE #372: Missing tab completion for commands
Weechat should inform glowing-bear about what commands are available and those should then work, so you can do /que<tab> and it completes to /query.

### LINKED ISSUE #1120: Nick and command completion
Very soon a new feature will be released in weechat. https://weechat.org/files/temp/doc/weechat_relay_protocol.en.html#command_completion

I'll probably do a pull request for this soon after the update is available. The command completion is new, but nick completion already existed and it is debatable if we should switch.

Arguments **contra** nick completion by this new command:
- Latency will reduce the feel of the app when used over high latency connections
- Nick completion would stop working for people using older weechat versions
  - Unless first check version, then choose which method to use, etc => more code

Arguments **pro** nick completion by this new command:
- Simplifies code if both nick and command completion uses the same mechanism and the old mechanism (searching the nicklist) is removed.
- Uses weechat's completion order so would fix #1063 
- Latency would only affect the first tab, after that cycling would be instantanious, unless new characters are input and tabbed again, however nicks are usually only one tab away

I plan to do the following:
- Use the new completion function for command completion #372 and anything following a command (command arguments), so whenever a line starts with '/'
- Leave nick completion as it is, using the nicklist
- If command completion works well, the latency is not an issue and weechat version 2.9 is used more, we can think about doing nick complation also.

Suggestion?

AStove:
A new command was introduced in 2.9 dev to be able to do completions of nicks and commands. Only implemented commands. https://weechat.org/files/doc/devel/weechat_relay_protocol.en.html#command_completion

- Don't do the old nick completion if line starts with /
- If line starts with / do command completion
  - Completes commands "/h" -> "/help, /halfop, ..."
  - Completes command arguments "/help " -> "/help -list, ..."
  - Cycle forwards with tab or the mobile button
  - Cycle backwards with shift-tab

fixes #1120
fixes #372

### DIFF
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -472,7 +472,7 @@ weechat.directive('inputBar', function() {
                     return true;
                 }
 
-                // Tab -> nick completion
+                // Shitft-Tab -> nick completion backward (only commands)
                 if (code === 9 && !$event.altKey && !$event.ctrlKey && $event.shiftKey) {
                     $event.preventDefault();
                     $scope.completeCommand('backward');
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -31,6 +31,9 @@ weechat.directive('inputBar', function() {
             // Emojify input. E.g. Turn :smile: into the unicode equivalent, but
             // don't do replacements in the middle of a word (e.g. std::io::foo)
             $scope.inputChanged = function() {
+                // Cancel any command completion that was still ongoing
+                commandCompletionInputChanged = true;
+
                 var emojiRegex = /^(?:[\uD800-\uDBFF][\uDC00-\uDFFF])+$/, // *only* emoji
                     changed = false,  // whether a segment was modified
                     inputNode = $scope.getInputNode(),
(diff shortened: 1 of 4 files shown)


AStove:
Some issues popped up, thanks @flashcode for testing:

- "/h" => "/help" instead of "/help " 
- "/quernick" (with cursor before "n") gives "/query" instead of "/query nick"
- if you press Tab to complete, and change the input before the response is received, it will complete anyway in your new input

flashcode:
For the third point, my suggestion is to just ignore the completion response if the user changed the input, and not change anything in the input.

AStove:
Fixed these problems.

js/inputbar.js
```diff
@@ -76,6 +79,11 @@ weechat.directive('inputBar', function() {
             };
 
             $scope.completeNick = function() {
+                if ( $scope.command.startsWith('/') ) {
```
lorenzhs, line 82: This breaks nick completion in situations like `/query foo<tab>` for weechat <= 2.8, we should probably check for weechat >= 2.9 as well before aborting. But we know the weechat version anyway (`models.version[0] >= 2 && models.version[1] >= 9`).

lorenzhs:
Wow thanks so much @AStove and @flashcode, this is absolutely fantastic. I'm going to set up a test weechat and play around with this a bit :)

js/inputbar.js
lorenzhs: (the above is wrong, I committed a fixed version)

lorenzhs:
This is absolutely perfect, I love it! Thanks so much guys!

PR #1124, opened 2020-04-30
Show toast with deletion link for imgur uploads

### LINKED ISSUE #1050: Uploading Images should ask consent
When sharing images with people glowing bear should really ask whether it is ok to upload those images to imgur by default. I was pretty surprised when it didn't ask me whether it's ok to upload the image but suddenly just showed me an imgur url. Especially from a GDPR perspective this seems kind of dangerous.
Could there be some way to ask for consent before? :)

AStove:
After uploading an image through pasting an image in, a toast is shown with a deletion link. On clicking the link the image is deleted. The hash is also shown in the log.
Changed the #toast css to .toast because it's used in multiple divs and they need unique id.

Things I'm not sure about:
- Would this qualify to fix #1050
- This doesn't cover uploading with the button, but that cannot be done accidental. Reason I don't do this is because this uploads multiple images, not sure how the deletion would work, etc.
- The thing I print in the console cannot be execute if not on a suitable breakpoint. Not sure how to do this better. Also I cannot just show a link because it's a DELETE call, not a GET. Is there some page I can link to on imgur that has the deletion link?

### DIFF
--- a/js/imgur.js
+++ b/js/imgur.js
@@ -104,10 +104,8 @@ weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, setting
         };
 
         if( "upload" in xhttp ) {
-
             // Set progress
             xhttp.upload.onprogress = function (event) {
-
                 // Check if we can compute progress
                 if (event.lengthComputable) {
                     // Complete in percent
@@ -117,12 +115,9 @@ weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, setting
                     currentProgressBar.style.width = complete + '%';
                 }
             };
-
         }
-
         // Send request with form data
         xhttp.send(fd);
-
     };
 
     // Delete an image from imgur with the deletion link
@@ -156,22 +151,16 @@ weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, setting
 
         // Handler for response
         xhttp.onload = function() {
-
             // Check state and response status
             if(xhttp.status === 200) {
-
                 callback();
-
             } else {
                 showErrorMsg();
             }
-
         };
-
-
+        
         // Send request with form data
         xhttp.send(null);
-
     };
 
     var showErrorMsg = function() {
--- a/js/imgur.js
+++ b/js/imgur.js
@@ -140,11 +140,6 @@ weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, setting
             isClientID = false;
         }
 
-        // Add the image to the provided album if configured to do so
-        if (!isClientID && settings.iAlb.length >= 6) {
-            fd.append("album", settings.iAlb);
-        }
-
         // Create new XMLHttpRequest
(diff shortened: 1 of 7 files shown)


js/inputbar.js
```diff
@@ -276,10 +277,21 @@ weechat.directive('inputBar', function() {
...
                         imgur.process(files[i], sendImageUrl);
                     }
-
                 }
             };
 
+            var deleteCallback = function () {
+                // Image got sucessfully deleted.
+                // Show toast with delete link
+                var toastDeleted = $compile('<div class="toast">Successfully deleted.')($scope)[0];
```
lorenzhs, line 286: that div isn't closed; also why are you using `$compile` when everywhere else we create elements without any angular involvement? I'd like to keep it consistent if there isn't a good reason to prefer `$compile`.

lorenzhs:
I worry that the 5s toast time is a bit short to realise what happened when pasting, maybe it should be 15s for this toast? I.e., split the animation specification out of the `.toast` class and moving it to new classes `.toastshort` and `.toastlong` with 15s delay or something? What do you think?

js/imgur.js
```diff
@@ -125,6 +125,55 @@ weechat.factory('imgur', ['$rootScope', 'settings', function($rootScope, setting
...
+            if(xhttp.status === 200) {
+
+                callback();
+
+            } else {
+                showErrorMsg();
+            }
+
+        };
+
```
lorenzhs, line 170: is your monitor like 8k or something or what's up with all the whitespace? ;)

AStove: Just a copy of ln 120, removed some, better?

js/inputbar.js
AStove: Closed the div. The reason for the complile is to make ng-click work. If I do a normal onClick I'd have to put the handler function somewhere in a random place, didn't want to do that. What do?

AStove:
That's why I also logged it in the console. There's no need to change the class, it can be changed for each toast, it's a setTimeout. Let's try 10? I mean when you paste something you realize very quick you made mistake.

js/inputbar.js
AStove: We could put the toast there permanently and then show and hide it.

lorenzhs: Nah sounds reasonable to use `$compile`, don't worry about it, I was just wondering.

lorenzhs:
Nope the actual toast hiding happens with a CSS animation that's hardcoded for a 0.5s fade-in, 4s active time, and 0.5s fade-out. The timeout is just to remove the element.

lorenzhs:
The problem now is that the "deleted" toast appears and then disappears, and the "wanna delete?" toast is visible again. We should probably add a data element with the deletion hash and then delete the matching toast...

lorenzhs:
Right, so that should work now, too. Anyone wanna test?

PR #1127, opened 2020-05-04
Remove support for non-native emoji rendering

lorenzhs:
It's not 2014 anymore, every reasonable platform supports displaying emoji natively these days. Let's remove a filter that has ceased to provide value.

### DIFF
--- a/css/glowingbear.css
+++ b/css/glowingbear.css
@@ -693,21 +693,6 @@ li.buffer.indent.private a {
     user-select: none;
 }
 
-.emojione {
-    font-size: inherit;
-    height: 1em;
-    width: 1.1em;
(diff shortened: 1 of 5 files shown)


lorenzhs:
Note that emojione is still used for text input of emoji (`:+1:` → :+1:) and thus not removed.

AStove:
Great, I was concidering updating that but removing it is even better.

AStove:
I think you forgot to remove the actual library from the index.hml

lorenzhs:
Nope I didn't forget, see the first comment above. It's still used to convert text-input emoji into actual emoji.

AStove:
Sorry I didn't read..

PR #1129, opened 2020-05-05
Delete unused library

AStove:
In previous PR #1127 this library should probably have been deleted.


lorenzhs:
See https://github.com/glowing-bear/glowing-bear/pull/1127#issuecomment-623986115

PR #1132, opened 2020-05-14
Fix unable to use history when input bar empty

### LINKED ISSUE #1130: Input history broken if nothing in input
It seems like the input history (arrow up) does't seem to work unless there's at least 1 character already in the input.

AStove:
Fixes #1130

### DIFF
--- a/js/inputbar.js
+++ b/js/inputbar.js
@@ -646,7 +646,7 @@ weechat.directive('inputBar', function() {
                         if ($scope.command) {
                             inputNode.setSelectionRange($scope.command.length, $scope.command.length);
                         }
-                    }, 0);
+                    }, 1);
                     return true;
                 }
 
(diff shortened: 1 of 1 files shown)


lorenzhs:
Hm is it intended that the cursor is in the first position when you scroll up from an empty line? I think it would be more natural if it was in the last position. I don't know whether that changed recently or not.

AStove:
It does go to the last position for me. Or do you mean on multiline inputs?

lorenzhs:
Nope it jumps to position 0 if I press the up arrow from an empty input bar. I'm using Firefox 76. Chrome works fine.

AStove:
That should be better. I've looked online a bit and there doesn't seem to be a better way to do this. Also I've seen someone do 1ms so they must have had the same problem.

lorenzhs:
Ugh. Thanks so much!

Issue #1055, opened 2019-07-10
Safari notifications and sounds are broken


lorenzhs:
Desktop Safari sounds suddenly stop playing sometimes. Notifications don't appear in Notification Center. These may or may not be related.

Reported by hiya on IRC.

lorenzhs:
`<hiya> lorenzh: So, the sound issue is fixed with auto-play "Allow All" in Pref - Websites - Autoplay`

fuzzy76:
Propose updating title and summary for this issue

lorenzhs:
Update to what? We still have no idea what exactly is going on with notifications in Safari, or whether it's possible to work around the auto-play thing, so for all intents and purposes, afaik, notifications and sounds are broken in Safari ;)

fuzzy76:
My bad, I thought the auto-play thing was evident. It's a feature in Safari 11 or newer to prevent web pages to play audio without user interaction. It should maybe be documented, but there is nothing glowing-bear can do to circumvent it.

Webkit team announcement: https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/
Reddit discussion: https://www.reddit.com/r/apple/comments/7165k8/safari_11_released_stops_media_with_audio_from/

So there is one bug here, and not two – namely the missing notifications. Which may be because of DoNotDisturb mode being on, that will mute notifications.

lorenzhs:
Thanks, so we should probably just add a notice to the setting that this won't work in Safari unless the user allows autoplay?

I'll check with `hiya` on IRC whether they might have had DND enabled.

AStove:
I think #1093 might be related.

AStove:
@fuzzy76 Would you know if this still happens? Latest release has an 'silent' audio file that plays upon connect. This would then allow subsequent auto play. This works in Chrome but I cannot check safari.

fuzzy76:
Unfortunately no, I don't use weechat any more. :-/ (I switched to thelounge since the web interface was more important to me than a shell client).

PR #1134, opened 2020-05-22
Add audio that plays on connect to enable autoplay

### LINKED ISSUE #1093: Highlight sound plays upon opening the tab
Often if there's a highlight while browsing in another tab the sound doesn't play. The favicon does change. The sound is then played upon opening the tab. Often but not always.

AStove:
fixes #1093 
(potentially #1055 could be fixed by this but I have not tested)

Implement a fix as described here: https://stackoverflow.com/questions/50490304/how-to-make-audio-autoplay-on-chrome

- Plays an mp3 with silence upon connecting
- Chrome now knows we have played audo and will allow autoplay later when a notification hits
- mp3 is as small as I could get it, 507byte. Potentially there might be smaller file possible. I tried with a very short .wav but this did not work
- Tested chrome and firefox
- Also works on autoconnect


lorenzhs:
Wow what a hack. Thanks.

AStove:
Thanks for the quick merge. Hope this solves it for good.

lorenzhs:
I just wanted to avoid adding another request, the file is so small that any loss of caching should easily be offset by reducing the number of requests (even though the overhead for that isn't as large as it used to be)

PR #1140, opened 2020-06-21
Improve Tik Tok Embed

AStove:
- Enable shortened links, since now finally their oembed accepts their own shortened links https://www.tiktok.com/oembed?url=https://vm.tiktok.com/JJnWEye/

- Make embed wider so desktop styling is applied. It's adaptive so it still goes back to mobile syling. The difference is of the bottons overlap the video or not. 

image


lorenzhs:
Cool, thanks!

PR #1141, opened 2020-06-28
Point link to https version of website

thomwiggers:
This URL linked via `http`, which means that it results in unnecessary warnings at the top of glowing-bear.org..


lorenzhs:
Thanks, I didn't know we still had an http link anywhere. That should have been changed long ago, thanks!

thomwiggers:
Is there any reason that the HTTP version doesn't redirect to the HTTPS version anyway?

lorenzhs:
Yeah, people using Glowing Bear on localhost can't easily use TLS, and you can't open an unencrypted websocket from an https page. It sucks, I know.

PR #1143, opened 2020-07-16
Use linkifyjs for links instead of linky

### LINKED ISSUE #961: Parentheses left out in URLs linkification
Many URLs have parentheses in them, especially those from wikipedia as for example https://en.wikipedia.org/wiki/Parenthesis_(disambiguation)

Right now, when this url is linkified in glowing-bear, the right parenthesis is left out (and therefore the link points to the wrong url)

AStove:
Wiki links often end in a parenthesis `)`, Linky doesn't include this in the link. This is a [known issue](https://github.com/angular/angular.js/issues/7213).

Since this would be rather hard to fix using linky I turned to a more modern library.
I also added a test.

Note, this library is in a newer javascript version, so I added a polyfill (mostly to make the unit tests work)

In a later PR I might move some of the bower components to node modules.

fixes #961


lorenzhs:
Thanks for looking into this, but isn't there another library that doesn't add 100kB polyfill + 45kB linkifier library? I get that the current linkification method isn't great (it bothers me, too) but I'm not sure we should add two network dependencies and 150kB of scripts for a rather minor improvement :/

lorenzhs:
Looking further, Autolinker's readme says that it compiles to ES5, so do we really need the polyfill? We don't support IE8 anyway: https://github.com/gregjacobs/Autolinker.js/#users-of-internet-explorer-8-and-below

AStove:
The Polyfill, we could get rid of, especially for the users, maybe not for PhantomJS, I could investigate why PhantomJS needs this.

Also we could probably get rid of angular-sanitize.min.js if we don't use Linky anymore but this is only 7kB gained.

We are talking about uncompressed kBs here, cached for up to a year. But yeah, I understand it's a slippery slope.

AStove:
Would you prefere this one? https://www.npmjs.com/package/linkifyjs It's 20kB.

lorenzhs:
afaik cache hits aren't really a thing on most mobile devices, the caches are too small and the sites people use too large to have a realistic chance of your stuff staying in cache. Maybe that info is outdated, though?

AStove:
Better?

22kB (new libs) - 7kB (removing ngSanitize) = 15kB

index.html
```diff
@@ -346,7 +347,7 @@ <h4 class="panel-title">
         </div>
 
         <div class="title"  ng-click="showModal('topicModal')">
-          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | linky:'_blank':{rel:'noopener noreferrer'} | DOMfilter:'irclinky')"></span>
+          <span class="desktop" ng-repeat="part in activeBuffer().title" ng-class="::part.classes" ng-bind-html="::(part.text | DOMfilter:'irclinky')"></span>
```
lorenzhs, line 350: why did you remove linkification from the topic?

lorenzhs: also, you don't need `DOMfilter` if you're passing plain text into it

AStove: Because I removed the old linky filter. I could have put conditionalLinky back in but upon testing I saw that links in topic worked..

AStove: I thought irclinky turns `#asdf` into channel links..

lorenzhs: yeah that's what `irclinky` does, but you don't need to wrap it in `DOMfilter` if you're passing `part.text` into it, `DOMfilter` is only needed if you pass the output of another filter that creates DOM elements

lorenzhs: Links in the topic "work" in pull.g-b because the updater is broken by the ES6 syntax in another recent pull request, it's still using angular's `linky` :see_no_evil:

AStove: Readded it, you are right, my testing method was wrong.

AStove:
Feel free to squash all these commits btw.

lorenzhs:
Now you're using angular's `linky` for the topic and the new one for the lines, should probably remove those occurrences of `linky` as well ;)

lorenzhs:
Also because you're using npm and not bower, the linkifyjs files aren't copied for electron.  You need to add a line
`cp node_modules/linkifyjs/dist/linkify{,-string}.min.js 3rdparty` to the `copylocal` target in `electron.makefile` and `npm install` to the `bower` target (or just use bower for now until we migrate everything at once, that's probably less overhead)

PR #1145, opened 2020-07-18
Downgrade rich text for buffers

### LINKED ISSUE #1043: Hidden Slack channels have strange name prefixes (square F02)
I just upgraded the wee_slack.py plugin, and in WeeChat everything looks fine. 

But in Glowing Bear the muted channels have a strange prefix (see screenshot).

I understand this might be more of a wee_slack.py problem but since this wasn't the case before I was wondering if this is some sort of style parsing issue in GB?

Screenshot 2019-04-11 at 18 56 31

AStove:
Applies the rich text function to buffer names so we can extract the raw text and filter out special characters for color and such.

fixes #1043

Could go further and apply the actual color.

### DIFF
--- a/js/models.js
+++ b/js/models.js
@@ -18,6 +18,10 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
     this.outgoingQueries = [];
 
     var parseRichText = function(text) {
+        if(!text) {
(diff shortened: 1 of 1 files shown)


lorenzhs:
Thanks!

PR #1146, opened 2020-07-18
Add link to make it easier to create an exception for self signed certificates

AStove:
If you want to use a self signed certificate you have to add an exception to be able to use it. To do this you have to visit the websoket location directly. Because browsers won't ask to add exceptions for 'subresources'. To make this accessible this PR adds a link in the warning to do this directly.

https://bugzilla.mozilla.org/show_bug.cgi?id=1187666

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,7 +64,7 @@
         <strong>Connection error</strong> The client was unable to connect to the WeeChat relay
       </div>
       <div class="alert alert-danger" ng-show="sslError" ng-cloak>
-        <strong>Secure connection error</strong> A secure connection with the WeeChat relay could not be initiated. This is most likely because your browser does not trust your relay's certificate. Please read the encryption instructions below!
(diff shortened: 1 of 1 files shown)


lorenzhs:
We used to have that under "Encryption Instructions" and removed it because letsencrypt makes it really simple to get a valid certificate. But even then it was hidden in the encryption instructions because we don't want to *encourage* people to add exceptions for untrusted certificates. It has the "SOMETHING HAPPENED AND YOU NEED TO PRESS OK TO GET ON WITH THINGS" popup vibe.

lorenzhs:
(also the URL is wrong, it needs to be `/weechat`)

AStove:
> (also the URL is wrong, it needs to be `/weechat`, or whatever the user entered into the host field if they entered the whole `host:port/location` string)

To accept the certificate exception, the "/weechat" part doesn't need to be there.

lorenzhs:
Yeah probably true but I wouldn't bet money that there isn't some way to create a messed-up setup where this matters

AStove:
But, you're right I guess.

PR #1148, opened 2020-07-21
Use linkifyjs instead of linky

AStove:
original pull request #1143 but needed to switch to a new branch.

Wiki links often end in a parenthesis `)`, Linky doesn't include this in the link. This is a [known issue](https://github.com/angular/angular.js/issues/7213).

Since this would be rather hard to fix using linky I turned to a more modern library.
I also added a test.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,7 +12,8 @@
     "angular-touch": "1.8.x",
     "angular-loader": "1.8.x",
     "angular-mocks": "1.8.x",
-    "underscore": "~1.10"
+    "underscore": "~1.10",
+    "linkifyjs": "jQuery-linkify#^2.1.9"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


AStove:
> Now you're using angular's linky for the topic and the new one for the lines, should probably remove those occurrences of linky as well ;)

Done

> Also because you're using npm and not bower, the linkifyjs files aren't copied for electron. You need to add a line
cp node_modules/linkifyjs/dist/linkify.min.js node_modules/linkifyjs/dist/linkify-string.min.js 3rdparty to the copylocal target in electron.makefile and npm install to the bower target (or just use bower for now until we migrate everything at once, that's probably less overhead)

Added to bower

lorenzhs:
Thanks a lot!

PR #1150, opened 2020-07-22
Omit getting script in oembed for twitter

AStove:
Omits the script from the request, making it so we don't have to trim it later.

Tested for TikTok but does not seem to listen to this argument.


lorenzhs:
Nice, works great for me!

lorenzhs:
Oh wait we manually append the script a few lines later, nvm

PR #1151, opened 2020-07-22
Pause audio, videos and iframe when hiding plugin

### LINKED ISSUE #975: Embeds: Videos, etc. are not automatically paused when collapsed
This is a small issue in the grand scheme of thing, but it would be a nice quality of life improvement.

The current behavior is that items such as YouTube videos and Streamable embeds will continue to play even when their button is toggled to close. It would be nice if they paused before collapsing.

AStove:
This tries to pause videos, audio. If it's an iframe (for example youtube or tiktok) it has to reload the iframe to get it to stop. 
Note this does not pause the videos in iframes such as youtube, so any progress in the video is lost, aswell als already downloaded data.

fixes #975

### DIFF
--- a/js/plugin-directive.js
+++ b/js/plugin-directive.js
@@ -31,6 +31,24 @@ weechat.directive('plugin', ['$rootScope', 'settings', function($rootScope, sett
 
             $scope.hideContent = function() {
                 $scope.plugin.visible = false;
+                // Pause noise makers
+                var element = $scope.plugin.getElement()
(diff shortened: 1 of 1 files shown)


lorenzhs:
Unloading the iframe could also be rather annoying, not sure whether it's better to keep it playing while collapsed or unload it and potentially having to reload it if it was a misclick.

lorenzhs:
Let's just see which behaviour annoys fewer people ;)

PR #1152, opened 2020-07-26
Prevent iOS lockscreen from showing the silent notification initializer

AStove:
Reported by @lorenzhs 

This is an attempt at mitigation, did not actually test. Probably hard to reproduce anyway.


lorenzhs:
Doesn't work for me (iOS 13.6) :/ The media info thing is still there when I switch tabs, switch apps, or disconnect. I have to close the tab to get rid of it.

PR #1155, opened 2020-08-02
Add start argument to youtube embeds if t arguement is present in url

### LINKED ISSUE #974: YouTube: links with time stamps are not embedded with the time stamp
Example video: https://youtu.be/q-Pb8QXRiWw?t=49

When clicking the button to embed, it only yields the URL `https://www.youtube.com/embed/q-Pb8QXRiWw?html5=1&iv_load_policy=3&modestbranding=1&rel=0`

AStove:
Checks for a `t=123` argument in the url. Did this in a seperate regex to not complicate the original. This is then added to the embed as a `start=123` argument.

Fixes #974

### DIFF
--- a/js/plugins.js
+++ b/js/plugins.js
@@ -216,9 +216,9 @@ plugins.factory('userPlugins', function() {
             var token = match[1];
 
             var matchTimeArgumentNoUnit = url.match(/(?:\?|\&|\#)*?(\d+)(?:$|\&)/i);
-            var matchTimeArgumentSecs = url.match(/(?:\?|\&\#)*?(\d+)s/i);
-            var matchTimeArgumentMin = url.match(/(?:\?|\&\#)*?(\d+)m/i);
-            var matchTimeArgumentHour = url.match(/(?:\?|\&\#)*?(\d+)h/i);
+            var matchTimeArgumentSecs = url.match(/(?:\?|\&|\#)*?(\d+)s/i);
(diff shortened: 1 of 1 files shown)


lorenzhs:
The argument for `t` can also be something like `1h23m42s`, so you should check for  `t=((\d+h)?(\d+m)?(\d+s?)?)` or something like that: https://www.youtube.com/watch?v=dQw4w9WgXcQ&t=1m42s

AStove:
This should do it.

lorenzhs:
Does the embed not support the same time formats, or why don't you just preserve the time argument as it was? Because on first look, this looks more complicated than it should need to be, and I'm wondering why the regex I proposed wouldn't work

AStove:
The `start` argument only takes seconds.
I'll give the regex another go using your suggestion.

lorenzhs:
You can capture all the groups in one regex using something like `t=(?:(\d+)h)?(?:(\d+)m)?(?:(\d+)s?)` and then group 1 is hours, group 2 is minutes, and group 3 is seconds. See regexr.com/5aad2.

AStove:
https://regexr.com/5ablc my tests
The main reason I didn't do this from the start wast hat I though order didn't matter, but it does, youtube doesn't allow t=5s5h for example.

lorenzhs:
Thanks! Looks good now :)

Issue #1078, opened 2019-11-29
Migrate from AngularJS to Angular (AngularJS End of Life)


baldurmen:
As far as I understand, this project depends on AngularJS, which has been End of Life since July 2018: https://www.convective.com/angularjs-end-of-life/

AngularJS is currently in Long Term Support until June 30 2021 and this project should move to an alternative, lilke [Angular](https://angular.io/).

This is likely a large shift, but I think starting now would be good idea :)

lorenzhs:
I don’t think anyone is prepared to put in the required amount of work. Angular is quite different from AngularJS, and the latter won’t stop working. It’s not great, but I don’t think a switch is likely.

baldurmen:
>  and the latter won’t stop working

I respectfully disagree on this. I don't think it's safe or responsible to tell people to use frameworks that have been deprecated and won't receive security updates.

> I don’t think anyone is prepared to put in the required amount of work.

Yeah, that I understand and it's totally OK to say you won't do that work :)

I am not familiar enough with Glowing Bear's code base, but how much work do you estimate that would be? If no contributor is ready to put in that amount of work, maybe we could try to crowdfund it? I'd be happy to give a few bucks if that would help migrate in time for 2021.

FWIW, [I'm currently working on packaging Glowing Bear (not the electron side) in Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=921707) and it certainly won't make it in the next stable release, since AngularJS won't make it either :(

lorenzhs:
> I respectfully disagree on this. I don't think it's safe or responsible to tell people to use frameworks that have been deprecated and won't receive security updates.

Sure, but what are we going to do in the absence of people willing to put in the work? Delete the repository to prevent people from using it?

> I am not familiar enough with Glowing Bear's code base, but how much work do you estimate that would be? If no contributor is ready to put in that amount of work, maybe we could try to crowdfund it? I'd be happy to give a few bucks if that would help migrate in time for 2021.

It's not about money, it's about time and motivation. Instead of porting our codebase, a partial rewrite would probably be the better approach. There's a lot of accumulated cruft (for example around scrolling, connection management, and swipe gestures) that shouldn't be allowed to survive a major restructuring such as porting to a different framework ;)

[...]

PR #1166, opened 2020-11-28
convert app to be built using webpack

yeiniel:
this is the first step on the roadmap to upgrade to the new version of the angular framework. It contributes to issue #1078 

the following changes are included:

 - add webpack to bundle app code
 - transform most of the javascript files into modules
 - upgrade travis to stop using a deprecated ubuntu image
 - upgrade karma to stop using a deprecated browser
 - remove bower and moved bower dependencies to npm


lorenzhs:
Cool, thanks! Unfortunately I'm a bit unqualified to review this, I've never used webpack and know absolutely nothing about it. It'd be great if someone else could have a look.

Am I correct though that this makes webpack a build step? I always liked that there was no build step involved in hosting glowing bear, you could simply clone the repo, modify a file, and reload the page in your browser to test it. I kinda liked that. But it seems that web development just doesn't work like that any more.

cormier:
Hi @yeiniel, thanks for your work ! This looks great. I don't have as much time as I would like this week but I will give it a look and give you feedback as soon as possible.

yeiniel:
> Cool, thanks! Unfortunately I'm a bit unqualified to review this, I've never used webpack and know absolutely nothing about it. It'd be great if someone else could have a look.
> 
> Am I correct though that this makes webpack a build step? I always liked that there was no build step involved in hosting glowing bear, you could simply clone the repo, modify a file, and reload the page in your browser to test it. I kinda liked that. But it seems that web development just doesn't work like that any more.

Hi @lorenzhs , yes you are right, it makes webpack the build step.

yeiniel:
`Am I correct though that this makes webpack a build step?`
Yes, that's the case. In the future during the upgrade we will need the legacy code to coexist with the new one and both to be packed together. Therefore doing the packing before helps on that process. Additionally we will need to split the code into smaller chunks to comply with angular style guide (one file per component or service rule) and easy the conversion to typescript. Without the packing having so many javascript modules becomes something ugly.

yeiniel:
Sorry I was trying to comment from my phone and closed the pr by mistake, I re opened it again

yeiniel:
>  I always liked that there was no build step involved in hosting glowing bear, you could simply clone the repo, modify a file, and reload the page in your browser to test it

by the way with the introduction of webpack i included the development server, the new NPM `start` script launch it. It reloads the browser for you every time a change to the source is made.

cormier:
I have reviewed the PR. It has been a while since I have looked this closely at the code of the project. ☺

Thanks @yeiniel for submitting this PR. Here are my comments:

Regarding the MR in general: 
I like this MR because it tackles the underlying issue of updating some libraries (npm and travis) and ways of doing things (using only npm for dependencies instead of bower + npm).

Regarding packaging and the build step: 
This PR introduces a build step. GB was built with the idea that one should just be able to clone the repo and be able to start using it. As @yeiniel has mentionned, we can use the webpack development server autoreload feature to have a simple development workflow.

I am in favour of introducing a build step and believe that overall it would bring more benefits than problems.
- Our CSS is large and hard to maintain, and I believe that adding a build step would make the integration of a CSS preprocessor like Sass easier.
- We could also add a step to minify the JS and CSS
- For the users who do not want to build GB, we could provide a package. This package could be built by travis on every tag. 

If we all agree in introducing the build step, I believe we should modify the following before merging this PR:

- We have a two "improve unit testing" commits and one "try to fix test exec on travis env". Can we squash these and add more details on _how_ it improves unit testing?
- The underscore import is commented out in 330216b59cd1c3bace0e1890645dab97fddab172. Can we remove it instead ?
- The server has to be started with `npm start`. Can we update the readme to reflect this ?

yeiniel:
i already made the suggested updates in case you all agree with merging the PR

lorenzhs:
Thanks for reviewing this, @cormier! I like the idea of adding a CSS preprocessor etc. later. I’m good to merge this once your concerns are addressed.

lorenzhs:
So are we good to merge this?

cormier:
We are !

lorenzhs:
Thanks again, @yeiniel! This is great.

PR #1170, opened 2020-12-13
Fix tab completion

### LINKED ISSUE #1169: Tab completion is not working
Getting

```
angular.js:15697 ReferenceError: _ is not defined
    at irc-utils.js:28
    at Object.completeNick (irc-utils.js:125)
    at f.t.completeNick (inputbar.js:106)
    at f.e.handleKeyPress (inputbar.js:522)
    at fn (eval at compile (angular.js:16548), <anonymous>:4:240)
    at i (angular.js:29123)
    at f.$eval (angular.js:19523)
    at f.$apply (angular.js:19622)
    at HTMLBodyElement.<anonymous> (angular.js:29127)
    at It (angular.js:3891)
```

whenever I try to tab-complete.

See https://github.com/glowing-bear/glowing-bear/blob/546b5d44598d7caa6c902b191af5cb0b97d04397/src/js/irc-utils.js#L28 for the source of the error. Seems like ~~lodash~~ underscore isn't being included due to #1166.

mcpower:
Fixes #1169.

Note that this doesn't use `import * as _ from "underscore";` - instead, it only imports the functions it needs for [tree shaking](https://webpack.js.org/guides/tree-shaking/) purposes. This doesn't help that much when done by itself, but I'm hoping that this is done in other files to get the full benefits of tree shaking.

### DIFF
--- a/src/js/irc-utils.js
+++ b/src/js/irc-utils.js
@@ -1,8 +1,8 @@
 /**
  * Portable utilities for IRC.
  */
+import {sortBy, pluck} from "underscore";
 
-(function() {
 'use strict';
 
 var IrcUtils = angular.module('IrcUtils', []);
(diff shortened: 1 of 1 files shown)


cormier:
@mcpower thanks ! this make sense. If I understand correctly, the idea of tree shaking is to import only what is needed so that the tree only contains "living code" in order to make the most efficient bundle possible ?

mcpower:
Yep - tree shaking is "dead code removal" for JS trees. It's particularly useful for large dependencies - Underscore seems to be around 8kb gzipped, so it's not _that_ important for Underscore, but it will definitely come in useful especially when migrating to Angular / another framework.

PR #1174, opened 2021-01-01
Improve Electron Support

rany2:
* Add a tray and close to tray on window close
* Add a context menu for copy image, copy, cut, paste, etc.
* Add a way to pass options to Electron app. Useful to hide on startup, set a custom profile, and enable devtools
* Disable dragging of HTML objects to make the app seem more native
* Disable selection of text when app is connected except for buffer (in order to make it behave more like a regular IRC client)
* Disable click and middle click on server list, etc. from opening up a new tab which serves no purpose
* Disable context menus on server list and nick list which shouldn't have context menus
* Enable keyboard shortcuts such as Ctrl+Q and others 
* Enable spellchecker, dictionary, and others 
* Fix icon and title 
* Fix init.json being created in current working directory instead of being created in `userData` 
* Fix webpack not installing audio files
* Force sandbox on all browser windows for security https://www.electronjs.org/docs/api/sandbox-option


oakkitten:
in some files there's mixed indentation (tabs + spaces), is that right?

also same indentation here https://github.com/rany2/glowing-bear/blob/18e566151303c023895c860a552de0d703f7a5f3/src/context-menu.js#L47

also the last commit has a weird message

rany2:
@oakkitten I saw a lot of mixed indentation before so I thought it didn't matter much. I'll fix styling now then

src/main2.js
```diff
@@ -0,0 +1,19 @@
+import "./js/localstorage.js";
```
lorenzhs, line 1: what's the reason for splitting into main and main2?

lorenzhs:
Thanks, @rany2! I'm just a bit confused because this pull request does a lot of unrelated things. Could you split it into separate PRs for things that don't have anything to do with electron, like the channel regex thing or making things unselectable? Having them all mixed together makes it harder to review than it needs to be.

Also, pinging @torhve for all things electron :)

src/main2.js
rany2: If you run `npm run build` webpack says you should split the code because the generated javascript ends up being to big. It just to solve that warning

rany2:
I'll do that soon :+1:

src/main2.js
yeiniel: that is because we are not compressing the output javascript. that would be the simplest solution

rany2: @yeiniel  So code splitting doesn't do much, right?

yeiniel: Code splitting can have a noticeable impact on the app loading but right now the reason for the message is just that we are not compressing the bundle.

PR #1186, opened 2021-04-11
Use self-hosted css and js

rain0r:


lorenzhs:
Thanks! When trying this out, I noticed that your changes to `package-lock.json` seem to include a local mirror, which doesn't resolve for anyone else and causes `npm install` to error out (running `npm update` gets rid of that, but of course does update other packages, too).

I'm a bit concerned that this blows up the size of the main.js bundle from ~300kB to 1.72MB. Does it somehow include all the CSS resources in the javascript, too? That would explain why the hostname/password form looks wrong (stylesheet loading order is changed). We probably shouldn't do that, and instead just copy the pre-minified css file from the npm package to our build folder and include it from there.

Also maybe we shouldn't include emojione in our bundle, it's huge (664kB minified), so if we just copy that to the build folder we can probably get better caching. Unless we can get the tree shaking thing to work with emojione, that would be neat (but probably a lot of the file size is for data, not code).

rain0r:
> Thanks! When trying this out, I noticed that your changes to `package-lock.json` seem to include a local mirror, which doesn't resolve for anyone else and causes `npm install` to error out (running `npm update` gets rid of that, but of course does update other packages, too).

You're right. Sorry for that - I fixed it.

> I'm a bit concerned that this blows up the size of the main.js bundle from ~300kB to 1.72MB. Does it somehow include all the CSS resources in the javascript, too? That would explain why the hostname/password form looks wrong (stylesheet loading order is changed). We probably shouldn't do that, and instead just copy the pre-minified css file from the npm package to our build folder and include it from there.
> 
> Also maybe we shouldn't include emojione in our bundle, it's huge (664kB minified), so if we just copy that to the build folder we can probably get better caching. Unless we can get the tree shaking thing to work with emojione, that would be neat (but probably a lot of the file size is for data, not code).

Yes, the bundle was pretty bight. Actually I couldn't figure out why exactly, because `bootstrap.min.css` is only 119K but ended up with ~700K in the bundle. But anyway, I followed your suggestion and now they will be loaded from the build dir.

Let me hear what you think!

rain0r:
Resolved the conflicts in package-lock.json.

rain0r:
@lorenzhs 
Is there anything I can do to get this merged?

PR #1189, opened 2021-05-26
Add a warning about freenode

lorenzhs:
Is this too much?


lorenzhs:
Given the feelings on our channel, I've merged this even though it's a bit late now. We can revert it again in a few months maybe. Right now, this only affects https://latest.glowing-bear.org/ and not the main subdomain.

PR #1190, opened 2021-05-28
Add Alt-J keyboard binding info

N6UDP:
Add Alt-J keyboard binding info


lorenzhs:
Thanks! Somehow we forgot about adding that :)

PR #1193, opened 2021-06-07
Strip protocol information from host field, checking for encryption

### LINKED ISSUE #1192: Unable to connect with HTTPS address, but port 443 works fine
A little oddity I discovered today when trying to connect to my relay. Entering https://relay will give an error about the browser not being able to trust the certificate. Or specifically "Secure connection error A secure connection with the WeeChat relay could not be initiated.".

Taking out the https:// part of the address and putting port 443 directly allows me to connect without problems.

lorenzhs:
Fixes #1192

### DIFF
--- a/src/js/glowingbear.js
+++ b/src/js/glowingbear.js
@@ -663,9 +663,20 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         $rootScope.hostInvalid = false;
 
         var parts;
+        var regexProto = /^(https?|wss?):\/\/(.+)$/;
         var regexHost = /^([^:\/]*|\[.*\])$/;
         var regexHostPort = /^([^:]*|\[.*\]):(\d+)$/;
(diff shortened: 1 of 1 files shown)


src/js/glowingbear.js
```diff
@@ -663,9 +663,20 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
         $rootScope.hostInvalid = false;
 
         var parts;
+        var regexProto = /^(https?|wss?):\/\/(.+)$/;
         var regexHost = /^([^:\/]*|\[.*\])$/;
         var regexHostPort = /^([^:]*|\[.*\]):(\d+)$/;
-        var regexHostPortPath = /^([^:]*|\[.*\]):(\d*)\/(.+)$/;
+        var regexHostPortPath = /^([^:]*|\[.*\]):(\d+)\/(.+)$/;
```
lorenzhs, line 669: We have to fix a little bug in the host:port/path regex to prevent "wss:/foo.bar" from being turned into host `wss`, port `<empty string>`, path `foo.bar` by requiring at least one digit between `:` and `/`

PR #1197, opened 2021-08-09
Revert "Change handshake timout to 200ms"

ericonr:
This reverts commit 5fb51c07e6a5bd2777c2cf4a7ae300419878c5b5.

200ms is too little for some setups, and will make authentication not
work even with weechat>=3.0.

For a specific example, my VPN setup has a roundtrip of 300ms.

### DIFF
--- a/src/js/connection.js
+++ b/src/js/connection.js
@@ -51,7 +51,7 @@ export const connectionFactory = ['$rootScope', '$log', 'handlers', 'models', 's
                     // too high, this time should be reduced if determined the weechat
                     // is lower than 2.9
                     // This time also includes the time it takes to generate the hash
-                    var WAIT_TIME_OLD_WEECHAT = 200; //ms
+                    var WAIT_TIME_OLD_WEECHAT = 2000; //ms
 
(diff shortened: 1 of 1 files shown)


Mikaela:
Since the change this PR reverts, I have been entirely unable to use Glowing Bear from home. I generally am connected to a VPN on desktop, which is too slow and it appears that my 4G connection is also too slow as I cannot connect even without a VPN from iPhone connected to 4G WiFi router.

Techman:
Yes, I hope this change gets reverted. Honestly, I'm surprised that the handshake isn't somewhere near 5 seconds. In addition to network latency, you never know if a host system has limited resources and if using higher strength hashing can introduce noticeable delay. Single-core VPSs come to mind.

This would explain why I essentially cannot connect using Firefox sometimes, but can at others. This is quite a frustrating experience to deal with as I essentially have to SSH/Mosh into my server to chat instead.

lorenzhs:
Let's see if this causes other people to complain but I'd say 2s of waiting is less bad than not being able to connect at all

PR #1198, opened 2021-08-26
Set attribute dir="auto" on message text span

amiryal:
Has no effect on non-RTL messages, but **helps greatly** with reading ones written in RTL languages, especially when intermingled with LTR text.

See [the spec](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute) for some hard-core technicalities, or scroll down to the example there for a visual demonstration.


lorenzhs:
Thanks! Let's see if this helps people.

PR #1205, opened 2022-01-24
Update webpack.config.js

jtracey:
This fixes broken dev builds introduced by 11d53f1.
See: https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md


lorenzhs:
Thanks and sorry for the slow response!

PR #1210, opened 2022-03-03
Add initial Tauri support

ogarcia:
This PR adds initial support for building the application with [Tauri](https://tauri.studio/).

[Tauri is a toolkit](https://tauri.studio/docs/about/intro) like Electron that builds web applications using much less disk space and much less memory.

I am not a node developer so I can't get the most out of the tool, the only thing I have done was an initial configuration that allows to generate a glowing-bear binary that takes 11MB and works correctly.

To generate binary you need install [Tauri dependencies](https://tauri.studio/docs/getting-started/setting-up-linux) and siply run:
```
npm run build
npm run tauri build
```

After end, the `glowing-bear` is located in src-tauri/target/release


src-tauri/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "app"
+version = "0.1.0"
+description = "A Tauri App"
```
lorenzhs, line 4: Can you fill in the name and a description?

src-tauri/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+authors = ["you"]
+license = ""
```
lorenzhs, line 6: GPL v3

lorenzhs:
Where does the web engine come from with tauri? Is it just using a shared one that isn't bundled? In that case, what's the advantage besides an apparently smaller size on disk? Where do the memory savings come from?

src-tauri/Cargo.toml
ogarcia: Ok. In Cargo.toml the format is [described here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields), in this case `GPL-3.0`.

lorenzhs:
Thanks! Let's see if people like it :)

Issue #1157, opened 2020-08-06
Critical bug: authentication method for weechat 2.9 does not work in http context


JeremyMahieu:
`crypto.subtle` or `window.crypto.subtle` is undefined in non secure contexts.

This is used for calculating the hash to log in to the new version of weechat with encrypted authentication.

Suggest either rolling back or disabling http. Don't really see a way around except with using an extra crypto library.

![image](https://user-images.githubusercontent.com/23409814/89577303-c9916400-d830-11ea-8cfa-6f1449d4a041.png)

AStove:
Maybe have an alert that shows when trying to connect to >=2.9 and crypto.subtle is undefined?

lorenzhs:
Well that's annoying. I don't know whether we should just say "GB doesn't work when served over http any more". There aren't a lot of valid use cases that require the http version (except for connecting to a different host on the same local network which can't use TLS). But maybe it would be better to fall back to the old insecure authentication when the site is used over http? But then again, showing big nasty warnings doesn't actually help (https://twitter.com/PowerDNS_Bert/status/1171352665164529665). Maybe disabling it is actually the way to go.

AStove:
The problem I have with falling back is, pre weechat 2.9 we don't even check if weechat accepts plaintext, because it always did. The result is regardless of weechat version: "Oh I see I'm in http, let's send the password in plaintext." So because you're in an insecure context, you end up being double insecure.

What would be reasonable imo is to fall back only when the compatibilty checkbox is on.

lorenzhs:
Yes, it's clearly a bad situation. The question is whether we want to support the very limited use case for that (-> add fallback, but without a compatibility checkbox because it's pointless) or not (-> kill http). I don't see the point of an in-between solution.

JeremyMahieu:
Visit glowing bear on http, and uncheck encryption.

Tony763:
Tried:
![image](https://user-images.githubusercontent.com/5598750/117936430-214f8500-b305-11eb-9573-118cbc65dd1c.png)
and
![image](https://user-images.githubusercontent.com/5598750/117936542-4512cb00-b305-11eb-8309-1d41f4c20c30.png)

Both result in:
![image](https://user-images.githubusercontent.com/5598750/117936470-2ca2b080-b305-11eb-89fd-fe58b057d67c.png)

lorenzhs:
> Visit glowing bear on http, and uncheck encryption.

[...]

PR #1212, opened 2022-03-29
Bug 1157 - Authentication in Non-Secure Context

tbabik:
PR with possible fix for bug report https://github.com/glowing-bear/glowing-bear/issues/1157
./run_tests.sh was executed and all unit tests were successful.

As it was discussed in the bug report, the downside is that it will again allow plain text password transmissions and even so there is a big red warning about it, some users may not understand the implications or security risks.

Index.html
- Changed index.html such that it shows an explicit danger alert when used in insecure context to alert the user that the password will be transmitted in plain text
- Changed user messages in index.html a little to ensure that the implication of compability mode with Weechat <2.9 also means sending plain text password (even if the user does not follow the link to the further information)

Connection.js
- Changed the connection logic such that if the handshakes timesout we just inform the user that they may be working with Weechat <2.9 and may need to use the Compatibility Mode; we do no longer make the assumption for them
- When using the compatibility mode the handshake is skipped and plain text password is exchanged
- When using it via insecure context handshake will be executed trying to agree on plain text exchange (will show an error if Weechat was configured to not support plain text)
- When using it via secure context handshake will be executed and will go for pbkdf2+sha512

webpack.config.js
- Debatable - I somehow saw no reason to be warned about the bundle size - but I guess there might also be more useful hints shown and would understand if this is reverted

glowingbear.js
- An additional variable to show the additional warning on index.html - but it is not so different from the above show_tls_warning, maybe it can be combined

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -36,7 +36,6 @@
         <small>WeeChat web frontend</small>
       </h2>
       <div class="alert alert-warning" ng-show="show_tls_warning" ng-cloak><strong>You're using Glowing Bear over an unencrypted connection (http://). This is not recommended!</strong> We recommend using our secure hosted version at <a href="https://www.glowing-bear.org/">https://www.glowing-bear.org/</a>, or <a href="https://latest.glowing-bear.org/">https://latest.glowing-bear.org</a> for the latest development version. If your relay is on your local network, that is unfortunately impossible, but be aware of the implications.</div>
-      <div class="alert alert-danger" ng-show="!isSecureContext" ng-cloak><strong>You're using Glowing Bear over an unencrypted connection.</strong> Password will be transmitted in plain text!</div>
       <div class="alert alert-danger" ng-show="errorMessage" ng-cloak>
         <strong>Connection error</strong> The client was unable to connect to the WeeChat relay
       </div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -117,7 +118,7 @@ <h4 class="panel-title">
                   <div class="checkbox">
                     <label class="control-label" for="compatibilityWeechat28">
                       <input type="checkbox" id="compatibilityWeechat28" ng-model="settings.compatibilityWeechat28">
-                      Compatibility with Weechat 2.8 and older <a href="#plaintext" ng-click="toggleAccordionByName('gettingStartedAccordion')"><i class="glyphicon glyphicon-info-sign"></i></a>
+                      Compatibility with Weechat 2.8 and older (Password in Plaintext) <a href="#plaintext" ng-click="toggleAccordionByName('gettingStartedAccordion')"><i class="glyphicon glyphicon-info-sign"></i></a>
```
lorenzhs, line 121: I find this confusing - when using TLS, the password is encrypted after all. It's being sent to the server, sure, but that's no different from every website login.

lorenzhs:
Apart from the wording on the option, I like this - haven't tested it but it looks great. Thanks a lot!

(Our unit test coverage isn't indicative of much, mostly plugin matching, so don't rely on that)

src/index.html
tbabik: True, I got carried away and seemed to have forgotten the possibility of using HTTPS + Pre 2.9.
I've removed the plaintext hint and also removed the warning panel, it does not really offer additional benefit compared to the existing "show tls warning" panel.

lorenzhs (approved):
:shipit:

lorenzhs:
Thanks a ton, this is great!

Tony763:
@tbabik thank You for Fix. Is there anything special needed to be set in weechat?

tbabik:
@Tony763 the Weechat setting that influences the allowed password algorithms is relay.network.password_hash_algo, but its default should bei "*" thus allowing both possibly needed algorithms "plain" and "pbkdf2 + sha512".

Besides that no configuration should bei required.

Tony763:
I works :tada: :tada: :tada:

Forget to rebuild app :sweat_smile:

PR #1216, opened 2022-04-18
Twitch plugin

moo-the-cow:
I renamed the branch from twitch-clips-plugin to twitch-plugin (because I added some more features)

Feel free to change button text or plugin position.
I added a technical reference for twitch embeds into the comments at the plugin method


lorenzhs (approved):
Cool, thanks!

PR #1218, opened 2022-05-08
add [] to the ignored end characters in url regex

jtracey:


lorenzhs:
What is this needed for?

jtracey:
It fixes a bug where a link to an image enclosed in brackets would successfully be parsed for display as a link, but the "Show image" button would fail to properly parse it. Try sending a message with
`https://glowing-bear.org/assets/img/glowing-bear.svg`
vs.
`[https://glowing-bear.org/assets/img/glowing-bear.svg]`
(in a glowing bear instance without this patch) to see what I mean. The regex already properly handles other delimiters, like `()` or `<>`.

lorenzhs:
OK, makes sense. Thanks. Our URL parsing is dumb as rocks anyway, so no harm in adding one more delimiter (it fails for URLs like `https://foo.bar/blah_(lol)`, where the last `)` is excluded - Wikipedia has lots of URLs like this. But `]` should be much rarer, so we should be good.

PR #1239, opened 2023-02-10
webpack: include sounds in webserver, and increase asset size so it

torhve:
does not warn for emojione and main.js


lorenzhs (approved):
Thanks!

PR #1243, opened 2023-02-12
Remove all corodova stuff

torhve:
We are never going to use cordova, best to remove everything


src/js/glowingbear.js
```diff
@@ -118,10 +111,10 @@ weechat.controller('WeechatCtrl', ['$rootScope', '$scope', '$store', '$timeout',
     })();
 
     // Show a TLS warning if GB was loaded over an unencrypted connection,
-    // except for local instances (local files, testing, cordova, or electron)
+    // except for local instances (local files, testing)
```
lorenzhs, line 114: I guess electron can remain here?

lorenzhs (approved):
Good riddance!

src/js/glowingbear.js
torhve: I will make a pull request to remove everything electron too

PR #1244, opened 2023-02-16
Remove underscore.js and reimplement all usage using native code

torhve:


src/js/models.js
```diff
@@ -219,15 +216,14 @@ models.service('models', ['$rootScope', '$filter', 'bufferResume', function($roo
...
-                _.each(nickGroup.nicks, function(nickObj) {
-                    if (nickObj.name === nick) {
-                        // Use the order the line arrive in for simplicity
-                        // instead of using weechat's own timestamp
-                        nickObj.spokeAt = Date.now();
+            for (let groupIdx in nicklist) {
+                let nicks = nicklist[groupIdx].nicks;
+                for (let nickIdx in nicks) {
+                    if (nicks[nickIdx].name === nick) {
+                        nicks[nickIdx].spokeAt = Date.now();
```
lorenzhs, line 223: I suppose we can break out of the loops after updating here?

lorenzhs (approved):
Pretty sweet! Thanks Tor!

src/js/irc-utils.js
```diff
@@ -19,17 +17,14 @@ IrcUtils.service('IrcUtils', [function() {
...
 
     /**
-     * Get a new version of a nick list, sorted by last speaker
+     * Get a new version of a nick list
      *
      * @param nickList Original nick list
-     * @return Sorted nick list
+     * @return list of nick names
      */
     var _ciNickList = function(nickList) {
```
lorenzhs, line 25: this is what's used for nick completion, are you sure we don't want to keep the sorting here?

lorenzhs: we could do `nickList.sort(sortBy('spokeAt')).map((el) => el.name).reverse()`?

lorenzhs: Ah, it's already filled with a sorted list (but we still need to reverse?)

PR #1246, opened 2023-02-17
Handle buffer cleared

torhve:
For some reason this was never implemented on our end - while I dont think its terribly important its nice to handle all events for completeness.


lorenzhs (approved):
Heh, good catch

PR #1257, opened 2023-06-02
Support for colors in buffer short name

torhve:
Took quite a few years before anyone tried that


lorenzhs (approved):
neat!

trygveaa:
Btw, as mentioned on IRC, if the short_name is changed while glowing bear is open, the color codes are shown rather than interpreted again.

Issue #1263, opened 2023-07-04
Desyncronized nicklist


JulienPalard:
I've not diagnosed the root cause (sooooory) but since a few time I notice some nicklist discrepency:

- Either it gets empty
- Either it gets way smaller than it should (in terms of number of participants)
- Either it's good (but I don't know, maybe it gets out of sync... ?)

Example:

![Capture d’écran du 2023-07-04 21-59-52](https://github.com/glowing-bear/glowing-bear/assets/239510/8179df81-0d69-4478-926b-eae971884ebd)

notice there's many people in the room but only the op displayed. I bet it is displayed because he's the last one changing or connecting.

Using Firefox 114, weechat-headless 3.8-1.

lorenzhs:
Autoclose was a bit aggressive there :upside_down_face:

jhoneycutt:
I can reproduce this by joining my bot to a channel I am in, kicking it, then rejoining it.

jhoneycutt:
Created PR #1274 to fix this.

lorenzhs:
Thank you so much! I think this will make quite a few people happy :) Merged!

PR #1274, opened 2024-02-02
Issue #1263: Desynchronized nicklist

jhoneycutt:
This is fallout from #1244, which moved models.delNick from using a filter to using the delete operator.

The delete operator creates a sparse array that has an empty slot for the item that was deleted. This empty slot is iterable with some constructs, giving a value of undefined. These unexpected undefined values in the nicklist cause exceptions to be thrown in some situations, like when calling models.updateNickSpeak.

To fix this, use Array.splice to remove elements from the nicklist, instead of the delete operator.


lorenzhs (approved):
:heart:

PR #1290, opened 2024-11-20
Upgrades from Tauri 1.0 to Tauri 2.0

ogarcia:
This change simply upgrades Tauri to 2.0 which is more modern and better.


torhve:
This PR probably makes sense, just haven't had the chance to properly look into it

torhve:
I also have this commit locally https://github.com/torhve/glowing-bear/commit/3d3377b6865495397f08eb5adda38df8e7bc36a0

to remove everything Electron

ogarcia:
> This PR probably makes sense, just haven't had the chance to properly look into it

No problem. Those of us who work in this field already know that there are only 24 hours in a day, and you can only do so much... :wink:

PR #1291, opened 2025-01-02
update references from ssl to tls to be in-line with weechat

tydel:
There are a few references to configuration parameters which no longer exist as they reference ssl rather than tls in versions >= 4.0.0 as per https://github.com/weechat/weechat/issues/1903


lorenzhs (approved):
thank you and sorry about the delay!

PR #1299, opened 2025-11-15
add docker image pushing to ghcr.io/glowing-bear/glowing-bear, and automated test runs

ferringb:
Per the title, this makes an OCI image available via your ghcr.io repo namespace.  This pushes `:latest` for changes to `master`, and pushes tagged versions for tags.

For sanity sake, that's wired to only run if tests pass.  Whilst you have some travis-ci configs, that looks dead- perhaps rewire to the GH actions instead?

A point of discussion is that I'm injecting a rust static file webserver in to the image to serve the resultant content- being you've got tauri, please double check if the static server thing I did is actually needed.  For some reason I recall tauri as being able to be a standalone web server.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+bower_components/
+node_modules/
+
+# Electron stuff
+fonts/
+Glowing\ Bear-*/
+
+# local build products
+build/ 
+
+# IntelliJ / WebStorm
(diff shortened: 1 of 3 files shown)


raghavmallampalli:
Can this be merged in? Would be nice to have an official docker

PR #1305, opened 2026-05-24
Cleanup: remove Libera banner, drop WeeChat 2.8 compat mode, update minimum version to 2.9

lorenzhs:
Six cleanup commits:

### Remove stale Libera Chat migration banner
The Freenode→Libera Chat migration happened years ago. Drops the dismissable banner and the `freenodeWarningRead` setting that tracked its dismissed state.

### Remove WeeChat 2.8 compat mode
WeeChat 2.9 was released in July 2020. The opt-in compatibility checkbox for the pre-2.9 plain-text init path is no longer warranted.

- Drop the 'Compatibility with Weechat 2.8 and older' checkbox and nested 'Use TOTP' sub-checkbox from the connection form
- Remove the compat section from the Getting Started docs
- Update the `oldWeechatError` message to just say "please update WeeChat"
- Remove `compatibilityWeechat28` / `useTotp` from settings defaults and the `useTotp` autoconnect callback
- Remove `parseTotp` / `totpInvalid` / `$scope.totp` (only existed to gate the Connect button in the compat-TOTP flow)
- Simplify `connection.js`: always perform the handshake; collapse the `weechatPre2_9` branches

### Update minimum WeeChat version to 2.9 everywhere
With compat mode gone, 2.9 is the hard minimum. Updated: `CLAUDE.md`, `README.md`, `src/index.html`, and two dead version guards in `glowingbear.js` / `models.js` that were unconditionally true at 2.9+ (dropped the `/query -noswitch` version check and the `getActiveBufferReference` else-branch).

### Fix plain-auth init to use 2.9+ wire format
The compat-mode removal left `_initializeConnectionPre29` handling the `"plain"` password path (non-secure http:// contexts), but it was sending the old `init password=…` format. WeeChat 2.9+, after a handshake, expects `init password_hash=plain:PASSWORD`. Fixed by replacing the call with `formatInit('plain:' + passwd, totp)`.

### Remove formatInitPre29, rename formatInit29 → formatInit
`formatInitPre29` (the old `init password=…` wire format) has no callers; deleted. `formatInit29` is now the only init formatter so the version suffix is dropped: renamed to `formatInit`.

### Rename _initializeConnection29 → _initializeConnection
Version suffix is no longer meaningful now that the pre-2.9 path is gone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/js/connection.js
+++ b/src/js/connection.js
@@ -92,27 +92,6 @@ export const connectionFactory = ['$rootScope', '$log', 'handlers', 'models', 's
             };
 
             // Helper methods for initialization commands
-            // Used for plain-text authentication (non-secure contexts)
-            var _initializeConnectionPre29 = function(passwd, totp) {
-                // Escape comma in password (#937)
-                passwd = passwd.replace(',', '\\,');
-
-                ngWebsockets.send(
-                    weeChat.Protocol.formatInitPre29({
-                        password: passwd,
-                        compression: noCompression ? 'off' : 'zlib',
-                        totp: totp
-                    })
-                );
-
-                // Wait a little bit until the init is sent
-                return new Promise(function(resolve) {
-                    setTimeout(function() { resolve(); }, 5);
-                });
-
-            };
-
-            // Helper methods for initialization commands
             // This method is used to initialize weechat >= 2.9
             var salt;
             var _initializeConnection29 = function(passwd, nonce, iterations, totp) {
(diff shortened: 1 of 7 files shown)


src/js/connection.js
```diff
@@ -45,49 +44,42 @@ export const connectionFactory = ['$rootScope', '$log', 'handlers', 'models', 's
...
-                    // 1. Compatability for Weechat 2.8 was activated by the user - skip handshake
-                    // 2. If SecureContext use pbkdf2+sha512 hash
-                    // 3. If !SecureContext use plain text
-                    // If handshake times out we do no longer make the assumption it is Pre 2.9 but just inform the user
+                    // If SecureContext use pbkdf2+sha512 hash, otherwise plain text
+                    // If handshake times out, inform the user their WeeChat is too old
 
-                    if (weechatPre2_9) {
-                        resolve();
+                    var WAIT_TIME_OLD_WEECHAT = 2000; //ms
```
lorenzhs, line 51: review with whitespace diff hidden

